PR #5, opened 2021-02-12
Improvements for finding the R executable

jonkeane:
and other efficiencies


R/ensure-lib.R
```diff
@@ -162,8 +162,8 @@ rspm_ids <- c(
 
 get_repo_url <- function(lib) {
   if (tolower(Sys.info()["sysname"]) == "linux") {
-    # TODO: bionic is hard-coded; use distro package to ID the OS properly
-    repo_url <-  paste0("https://packagemanager.rstudio.com/all/__linux__/bionic/", rspm_ids[lib])
+    # RSPM will send binaries as a fallback, so if the distro isn't perfect the worst that should happen is source installs.
```
nealrichardson, line 165:
```suggestion
    # TODO: non-Ubuntu OSes need different handling
    # RSPM will send binaries as a fallback, so if the distro isn't perfect the worst that should happen is source installs.
```

R/ensure-source.R
```diff
@@ -26,19 +28,27 @@ ensure_source <- function(file) {
     ext <- file_ext(known$url)
     file <- paste(data_file(file), ext, sep = ".")
     if (!file.exists(file)) {
-      utils::download.file(known$url, file, mode = "wb")
+      # override the timeout
+      # TODO: retry with backoff instead of just overriding?
```
nealrichardson, line 32:
```suggestion
      # TODO: retry with backoff instead of just overriding? or use `curl`?
```

nealrichardson (approved):
Two TODO comment suggestions; will apply them and merge

PR #8, opened 2021-02-22
Adjust the way that progress ticks to be ignored by default 

jonkeane:
(e.g. `run_one()`) and update how progress is polled


tests/testthat/test-run.R
```diff
@@ -1,3 +1,5 @@
+wipe_results <- function() unlink(test_path("results/placebo"))
```
nealrichardson, line 1: Arguably this should be a function in the package itself, seems useful

PR #9, opened 2021-02-22
Add df -> arrow table benchmark and chicago traffic dataset

jonkeane:


R/bm-R-to-arrow.R
```diff
@@ -0,0 +1,37 @@
+#' Benchmark file writing
+#'
+#' @section Parameters:
+#' * `source` A known-file id, or a CSV(?) file path to read in
+#' * `format` One of `c("parquet", "feather", "fst")`
+#' * `compression` One of `c("uncompressed", "snappy", "zstd")`
+#' * `input` One of `c("arrow_table", "data_frame")`
+#'
+#' @export
+df_to_table <- Benchmark("df_to_table",
```
nealrichardson, line 10: We should name the file `bm-df-to-table.R` to match the Benchmark name

R/bm-R-to-arrow.R
```diff
@@ -0,0 +1,37 @@
...
+  setup = function(source = names(known_sources), ...) {
```
nealrichardson, line 11: remove ... now?

R/bm-R-to-arrow.R
```diff
@@ -0,0 +1,37 @@
+#' Benchmark file writing
+#'
+#' @section Parameters:
+#' * `source` A known-file id, or a CSV(?) file path to read in
+#' * `format` One of `c("parquet", "feather", "fst")`
```
nealrichardson, line 5: These params don't seem to match, nor does the docstring title

nealrichardson (changes requested):

R/bm-R-to-arrow.R
jonkeane: Yes, thanks for catching that. I had removed all of them on the _other_ branch but not here

jonkeane:
This should be ready to go, any objections to merging?

PR #10, opened 2021-02-22
Add some CI goodness

jonkeane:


DESCRIPTION
```diff
@@ -22,10 +23,12 @@ Imports:
...
-  dplyr,
-  readr,
-  vroom
+    arrow,
+    data.table,
+    dplyr,
+    fst,
+    readr,
+    vroom,
+    covr
```
nealrichardson, line 32: Maybe don't add covr and only install it in the coverage job (or delete the coverage job)?

nealrichardson (commented):
One suggestion but otherwise LGTM

PR #11, opened 2021-02-23
Ensure that arguments aren't silently eaten + ignored

jonkeane:
This has bitten me a few times (and a few others) With this change, passing unknown arguments to a benchmark will fail. I didn't have to change much about how things are run, mostly removed the `...` from the `setup()` step of each (and had to remove our system/global level ones from the list before passing it to `setup()`


jonkeane:
I based this on #10 (so I could have GHA check that things were running correctly) so that should merge before this.

nealrichardson (approved):
I'm assuming you've confirmed that the actual benchmarks still run with this change

Issue #4, opened 2021-02-12
Framework for defining which benchmarks are Arrow


nealrichardson:
We have two related challenges:

* Some parameters we want to test are only relevant for Arrow. Some of these, like whether to return an arrow Table or data.frame when reading a file, may be specific to a single benchmark. Others, like environment variables that only affect Arrow, should be controllable at the global level. We don't want to run non-arrow benchmarks parametrized by arrow-only environment variables--that's wasteful/slow.
* When running benchmarks continuously (on every commit), we only want to run the Arrow benchmarks--no point wasting electrons on other people's (unchanged) code.

What if we added a function (`arrow_params()`?) similar to `default_params` that each benchmark can register that indicates which param combinations involve arrow? We can use that in whatever `run_all()` function we add for conbench, and we can also use it to only add variations for arrow env vars to the combinations that involve arrow.

PR #13, opened 2021-02-25
Add ability to install from hash

jonkeane:
Arrow is now installable from arbitrary remotes/branches by specifying the `lib_path` as `remote-apache/arrow@branch-name` This might be better with something like #4 since this will only work with arrow right now, but when I started down that road it started expanding scope into here so we might do this for now and circle back to that later.

One funny bit that we could work around, but I think it's fine is that the lib_path will have one extra level of hierarchy compared to the others because there is a `/` in the path. We can redact that when making the lib path, but I haven't run into any issue with that extra directory causing any issues.


R/ensure-lib.R
```diff
@@ -86,6 +90,31 @@ ensure_lib <- function(lib = NULL, test_packages = unlist(strsplit(packageDescri
...
 }
 
+install_arrow_github <- function(repo = "apache/arrow", ref = "HEAD", ...) {
+  # TODO: check if cmake is found? Say to use brew to install that?
+  # Do we want or need to also allow this to install any github package. For
+  # most it would only need to change the repo arg + expose the subdir argument
+  with_envvar(
+    list(
+      # always use the tools/linuxlibs.R script for installing arrow lib
+      USE_TOOLS_LIBS_SCRIPT = "true",
```
nealrichardson, line 100: This requires an arrow change right? We can 🚲 🏠  the name there.

R/ensure-lib.R
```diff
@@ -86,6 +90,31 @@ ensure_lib <- function(lib = NULL, test_packages = unlist(strsplit(packageDescri
...
+      # we want everything to be available
+      LIBARROW_MINIMAL = "false",
+      # Don't use pkgconfig because we don't want to find sources that are already installed
+      ARROW_USE_PKG_CONFIG = "false",
```
nealrichardson, line 104: I think you'll want the new "use script" flag to set this value in `configure` too, so you shouldn't need to set it here.

R/ensure-lib.R
```diff
@@ -86,6 +90,31 @@ ensure_lib <- function(lib = NULL, test_packages = unlist(strsplit(packageDescri
   lib_dir_path
 }
 
+install_arrow_github <- function(repo = "apache/arrow", ref = "HEAD", ...) {
+  # TODO: check if cmake is found? Say to use brew to install that?
```
nealrichardson, line 94: The build script will download cmake if it isn't found. You may need to revise that script to download a platform/arch appropriate binary though.

nealrichardson (commented):
Some comments but I think they're really for discussion on apache/arrow

R/ensure-lib.R
jonkeane: yeah, I went through a few others and this seems the least super-hacky but not super happy about it lemme send that as a PR and we can decide there before we merge this I'll update

R/ensure-lib.R
jonkeane: yeah, the one that is downloaded by default isn't right, but I will poke at that on the arrow side.

nealrichardson: macos has a different URL: https://cmake.org/download/

PR #17, opened 2021-03-11
Catch rlang style errors as well as base errors.

jonkeane:


R/measure.R
```diff
@@ -40,7 +40,9 @@ with_gc_info <- function(expr) {
   with_gcinfo <- "bench" %:::% "with_gcinfo"
   gc_output <- with_gcinfo(eval.parent(expr))
   # This will swallow errors, so check for error output and re-raise
-  if (length(gc_output) > 0 && startsWith(gc_output[1], "Error in ")) {
+  if (length(gc_output) > 0 &&
+      # catch both base "Error in " and rlang "Error: " style errors.
+      ( startsWith(gc_output[1], "Error in ") | startsWith(gc_output[1], "Error: ") )) {
```
jonkeane, line 45: Alternatively we could change this test to simply `startsWith(gc_output[1], "Error")`, but I presumed that there was worry of it being overly broad so I've added the rlang pattern specifically.

nealrichardson: I think startsWith("Error") is fine, gcinfo should never say "Error" on its own.

PR #18, opened 2021-03-11
Datasource standardization

jonkeane:


jonkeane:
When we actually started running the CI with tests, it looks like the subprocess on windows doesn't quite work. I've added a few skips there, which we should come back to (but not now)

R/ensure-source.R
```diff
@@ -1,62 +1,94 @@
...
+    # Check for places this file might already be and return those.
+    if (file.exists(data_file(filename))) {
+      # if the file is in our temp storage or source storage, go for it there.
+      return(data_file(filename))
     }
-    ext <- file_ext(known$url)
-    file <- paste(data_file(file), ext, sep = ".")
+
+    # Look up, download it
+    ext <- file_ext(filename)
```
nealrichardson, line 43: It doesn't look like you use `ext` anymore
```suggestion
```

R/ensure-source.R
```diff
@@ -1,62 +1,94 @@
...
+
+temp_data_file <- function(...) {
+  source_data_file("temp", ...)
+}
+
+#' Find a data file
+#'
+#' This looks in both the source dir ("data") as well as the temp directory
+#' ("data/temp") and returns the temp directory file if it exists, if it doesn't
+#' but the source dir exists it will return that, if neither exist it returns
```
nealrichardson, line 74: This behavior is confusing, can you explain why this is desirable?

R/ensure-source.R
```diff
@@ -180,17 +148,98 @@ ensure_dataset <- function(name, download = TRUE) {
...
+#' @param compression compression to be ensured
+#'
+#' @return the file that was ensured to exist
+#' @export
+#'
+#' @importFrom utils write.csv
+ensure_format <- function(
+  name,
+  format = c("csv", "parquet", "feather", "csv.gz"),
+  compression = c("uncompressed", "snappy", "zstd", "gzip", "gz", "lz4")) {
```
nealrichardson, line 176: I don't think you want both "gzip" and "gz" here. Too many places in the function check `if (compression == "gzip")` so would fail if you provide "gz". 

For better or worse, `match.arg()` will just partial-match `gz` to `gzip` anyway, so even if there were a good reason to support passing `gz` as an argument, it will still work.

```suggestion
  compression = c("uncompressed", "snappy", "zstd", "gzip", "lz4")) {
```

R/ensure-source.R
```diff
@@ -180,17 +148,98 @@ ensure_dataset <- function(name, download = TRUE) {
...
+    if (compression == "gzip") {
+      R.utils::gzip(csv_filename, ext = "gz", remove = FALSE)
+    }
+  } else if (format == "parquet") {
+    if (compression == "gzip") {
+      stop("gzip + parquet is not a known compression", call. = FALSE)
+    }
+    arrow::write_parquet(tab, sink = file_out, compression = compression)
+  } else if (format == "feather") {
+    if (compression == "gzip") {
```
nealrichardson, line 237: You also can't write feather + snappy. 

We seem to have this logic also in the valid_params methods for read_file and write_file benchmarks, so maybe we need a centralized function/object that tells you what file formats and compressions are valid combinations.

R/util.R
```diff
@@ -83,3 +83,17 @@ find_r <- function() {
 }
 
 is_macos <- function() tolower(Sys.info()["sysname"]) == "darwin"
+
+local_dir <- function() {
```
nealrichardson, line 87: Can you add a comment explaining the difference between local_dir and local_data_dir?

Do we need to support both env var and option?

R/ensure-source.R
jonkeane: It's more or less to match how the python benchmarks package works. Which is the flow:

* if a data source exists in `data/` use that
* if a data source exists in `data/temp` use that
* if there is no data source in either, you will need to end up needing to create it and put it in `data/temp`

I absolutely agree this function as it is is very odd — I was trying (probably too hard) to be semi-tactical with these changes and not change this internal API too much. But the more that I used it the more that I think there three functions aren't great.

Instead, what I think we should do (and I will do unless you have a better idea) is to have a function that returns the path in `data` or `data/temp` if it exists, and null if it doesn't. And then put the logic for "if this is `NULL` do the things that are needed to make the file exist" in `ensure_format()` (and a bit in `ensure_source()` too probably, though that will be a bit more limited). That way the fallback / search space logic is separated from the where-to-put-it logic and aren't both relying on code that returns paths that sometimes exist and sometimes do not.

R/ensure-source.R
jonkeane: Yeah, I only had it because I have the special case `.csv.gz` later and that was on my mind.

R/ensure-source.R
jonkeane: Yeah, I like the idea of keeping the valid options centrally. I hadn't (yet) pulled out the code in `read_file` and `write_file` that those use to do the transformation, but that was my plan for next steps was to pull that out and use the new `ensure_format()` and it would be good to use a single place to figure out what options are possible here and in the benchmarks.

FWIW, feather + snappy should never be triggered since the `write_file` and `read_file` should never ask for that, but I agree centralization for this is a better solution.

R/util.R
jonkeane: Sure, can do — we don't need to I left it in more for backwards compatibility (though that's probably silly for something this new). I'm happy to pull out the options if you think they are more confusing / not worth keeping around.

nealrichardson: Yeah I don't think you have to worry too much about breaking APIs, especially rather internal ones like this.

R/ensure-source.R
nealrichardson: Yeah I agree that sounds better

tests/testthat/test-util.R
```diff
@@ -1,4 +1,14 @@
...
   expect_identical(bm_run_cache_key("foo", alpha = "one", beta = 2), "foo/one-2")
   expect_identical(bm_run_cache_key("foo", beta = 2, alpha = "one"), "foo/one-2")
+})
+
+test_that("find_r()", {
+  out <- system(paste(find_r(), "--no-save -s 2>&1"), intern = TRUE, input = "print('output')\n")
+  expect_match(out, "output")
+
+  # the stderror isn't redirected correctly on windows, at least in GHA
+  skip_on_os("windows")
```
nealrichardson, line 11: Rather than skip on windows, I'd just comment windows out of the GHA build matrix and leave a note there (and make an Issue) that we will need to do work to support windows. Sounds like it fundamentally doesn't work on windows, so there's no point in papering over that by skipping tests.

PR #21, opened 2021-03-22
Remove extra compression checks

jonkeane:
And also set defaults to be the tests we actually care about running most of the time.


R/bm-read-file.R
```diff
@@ -3,18 +3,17 @@
...
 #' @export
 read_file <- Benchmark("read_file",
   setup = function(source = names(known_sources),
                    # TODO: break out feather_v1 and feather_v2, feather_v2 only in >= 0.17
-                   format = c("parquet", "feather", "fst"),
+                   format = c("parquet", "feather"),
                    compression = c("uncompressed", "snappy", "zstd", "lz4"),
                    output = c("arrow_table", "data_frame")) {
-    format <- match.arg(format)
-    compression <- match.arg(compression)
```
jonkeane, line 17: We could keep this line as `compression <- match.arg(compression, known_compressions)` but the compression is checked a few lines later inside of `ensure_format()` so I don't think it's necesary (and one less place to have it)

PR #22, opened 2021-03-31
Add how-to-write a benchmark to readme

jonkeane:


README.md
```diff
@@ -109,6 +103,124 @@ from `run_benchmark()` (both default `FALSE`):
   If `TRUE`, the result data will contain a `prof_file` field, which you can 
   read in with `profvis::profvis(prof_input = file)`. 
 
+## Defining benchmarks
+
+Benchmarks are constructed by `Benchmark()`, which takes expressions that handle 
+setup, teardown, and the actual work that we want to measure. See its 
+documentation for details, and see `read_file` and `write_file` for examples.
+
+### A (contrived) example
```
nealrichardson, line 112: Rather than adding a contrived version of the `read_csv` benchmark, what if you added these extensive comments to the actual benchmark and pointed to that from here? That way there's less risk that our detailed, documented example becomes stale.

README.md
```diff
@@ -109,6 +103,124 @@ from `run_benchmark()` (both default `FALSE`):
...
+  # This lists any packages that are used by this benchmark so that they can 
+  # be installed prior to starting the run. Typically this will be simply "arrow"
+  packages_used = function(params) {
+    "arrow"
+  }
+)
+```
+
+And now we could run our benchmark with the following for the default matrix, 
+using 4 cpu cores and 5 iterations per case.
```
nealrichardson, line 195: Should explain "iterations" somewhere, probably above under "Running benchmarks"

README.md
```diff
@@ -109,6 +103,124 @@ from `run_benchmark()` (both default `FALSE`):
...
+  as_data_frame = c(TRUE, FALSE),
+  skip_empty_rows = c(TRUE, FALSE),
+  cpu_count = 4,
+  n_iter = 5
+)
+```
+
+### Enabling benchmarks to be run on conbench
+
+Before a benchmark can be run on conbench, one must add a (or extend an existing)
```
nealrichardson, line 219: Link to conbench service (and later add link to blog post too). Also may want to start this paragraph with a sentence stating what conbench is and why you'd want to run them there.

README.md
jonkeane: Yeah, I had thought about doing that, but I worry that some of the uniqueness in that benchmark (and honestly each of them have their own quirks) would distract more than it would help. I will go back and comments to those benchmarks as well.

PR #24, opened 2021-04-15
Add benchmark for dim.Dataset

lidavidm:
In ARROW-9697 I'm working on a count rows method and want to make sure it doesn't regress performance (spoiler alert: it does). 

From what I can see, this also needs to be hooked into ursacomputing/benchmarks before conbench will pick it up.

### DIFF
--- a/R/bm-dataset-taxi-parquet.R
+++ b/R/bm-dataset-taxi-parquet.R
@@ -93,7 +93,7 @@ dataset_taxi_parquet <- Benchmark("dataset_taxi_parquet",
         dim(ds)
       },
       assert = function(result) {
-        stopifnot(identical(result, c(1547741381L, 20L)))
+        stopifnot("dims does not match" = identical(result, c(1547741381L, 20L)))
       }
     )
   ),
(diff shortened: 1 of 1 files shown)


dianaclarke:
Here's a PR to enable the new case after this merges.

https://github.com/ursacomputing/benchmarks/pull/29/files

lidavidm:
Thank you! I didn't realize this benchmark had already been mostly hooked up.

R/bm-dataset-taxi-parquet.R
```diff
@@ -87,6 +87,14 @@ dataset_taxi_parquet <- Benchmark("dataset_taxi_parquet",
           identical(sum(result$n), 3069271L)
         )
       }
+    ),
+    count_rows = list(
+      query = function(ds) {
+        dim(ds)
+      },
+      assert = function(result) {
+        stopifnot(identical(result, c(1547741381L, 20L)))
```
jonkeane, line 96: Minor, and you don't have to take it if you don't want, but this will provide a slightly more informative message if the assertion fails. (Also, yes this syntax is strange, don't ask me how it got to be this way)

```suggestion
        stopifnot("dims does not match" = identical(result, c(1547741381L, 20L)))
```

jonkeane (approved):
This looks great, thank you for submitting it! One small addition that you can take or leave.

lidavidm:
That's quite interesting syntax, but I'll take it - thanks!

PR #33, opened 2021-09-09
TPC-H benchmarks [WIP]

jonkeane:


R/bm-tpc-h.R
```diff
@@ -0,0 +1,206 @@
...
+  # packages used when specific formats are used
+  packages_used = function(params) {
+    c(params$engine, "dplyr")
+  }
+)
+
+# all queries take an input_func which is a function that will return a dplyr tbl
+# referencing the table needed.
+#' @export
+tpc_h_queries <- list(
```
nealrichardson, line 171: You could also do this as:

```r
tpc_h_queries <- list()
tpc_h_queries[[1]] <- function(input_func) ...
tpc_h_queries[[6]] <- function(input_func) ...
```

Could be more intuitive than stringified number names

R/bm-tpc-h.R
```diff
@@ -0,0 +1,206 @@
...
+        ensure_format,
+        FUN.VALUE = character(1),
+        format = format_to_convert
+      )
+
+      # specify readers for each format
+      if (format == "parquet") {
+        input_functions[["arrow"]] <- function(name) {
+          file <- tpch_files[[name]]
+          return(arrow::read_parquet(file, as_data_frame = FALSE))
```
nealrichardson, line 56: Why not open_dataset()? I wonder if that has different/better properties, even if it is a single file. (Hey we should benchmark that!) It at least shouldn't be memory constrained.

R/bm-tpc-h.R
```diff
@@ -0,0 +1,206 @@
...
+        }
+      } else if (format == "feather") {
+        input_functions[["arrow"]] <- function(name) {
+          file <- tpch_files[[name]]
+          return(arrow::read_feather(file, as_data_frame = FALSE))
+        }
+      } else if (format == "native") {
+        # native is different: read the feather file in first, and pass the table
+        tab <- list()
+        for (name in names(tpch_files)) {
```
nealrichardson, line 66: So we have to read in all tpch_files now? Do these queries share the same dataset, or do we get 22 datasets for 22 queries? 

Since each process is only running a single query (right?) it seems like you should be able to only read in the ones that the current query requires

.github/workflows/R-CMD-check.yaml
```diff
@@ -17,10 +17,11 @@ jobs:
             # TODO: figure out if this runs correctly on windows at all
             # - {os: windows-latest, r: 'release'}
             # We explicitly set the user agent for R devel to the current release version of R so RSPM serves the release binaries.
-          - {os: ubuntu-16.04,   r: 'devel', rspm: "https://packagemanager.rstudio.com/cran/__linux__/xenial/latest", http-user-agent: "R/4.0.0 (ubuntu-16.04) R (4.0.0 x86_64-pc-linux-gnu x86_64 linux-gnu) on GitHub Actions" }
-          - {os: ubuntu-16.04,   r: 'release', rspm: "https://packagemanager.rstudio.com/cran/__linux__/xenial/latest"}
-          - {os: ubuntu-16.04,   r: 'oldrel',  rspm: "https://packagemanager.rstudio.com/cran/__linux__/xenial/latest"}
-          - {os: ubuntu-16.04,   r: '3.5',     rspm: "https://packagemanager.rstudio.com/cran/__linux__/xenial/latest"}
+          - {os: ubuntu-20.04,   r: 'devel',   force_suggests: true,     rspm: "https://packagemanager.rstudio.com/cran/__linux__/xenial/latest", http-user-agent: "R/4.0.0 (ubuntu-16.04) R (4.0.0 x86_64-pc-linux-gnu x86_64 linux-gnu) on GitHub Actions" }
```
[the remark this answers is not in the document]
jonkeane: Yeah, thanks for pointing that out, I did catch it in the covr job, but should have also backported it here too.

R/bm-tpc-h.R
jonkeane: Ah, yes of course. This is much better

R/bm-tpc-h.R
jonkeane: I'll add that as an option and see if there are better properties. You're right we probably should always do `open_dataset()` here even if these are typically one-file datasets.

R/bm-tpc-h.R
jonkeane: That is what this is doing, yeah. Each query uses the same dataset (well dataset of multiple tables). This query only needs one table (lineitem), so we really only need to read that in.

I can make a query to table map so that we only read in the tables that are necessary for each query.

Ultimately, I don't think this will have all that much of an impact, lineitem is the largest table  by far, so reading the others in shouldn't have a huge impact even if they aren't used.

Additionally, right now what happens is that each of the files are memory mapped by this, but aren't actually paged into memory until the first run (this is why for arrow the first run in this condition is so much longer than the others). I'm planning to turn memory mapping off so that we really are starting with a truly in-memory table (though we could parameterize that if we wanted to!)

jonkeane:
Well, _it turns out_ `open_dataset("single_file.parquet")` is significantly faster than reading the file in (even as an Arrow table) and operating on it. I'm a little bit surprised (especially for the feather files!), but I wonder if the dataset scans of the files are more optimized than a query against a table that is backed by a file? 

I'm also pretty surprised that `open_dataset(feather_file)` is faster than the query engine against the table already resident in memory (the "native" format).

Any thoughts about an explanation for these oddities @bkietz?

With datasets, our performance against parquet files is in line with DuckDB's, though our native query processing is considerably longer.

Turning off memory mapping has the impacts I expected:

* For the native workflow, the first iteration is no longer considerably longer (since the file is already resident in memory before timing starts) 
* For the feather and parquet there's mixed results — here reading the file in happens during the benchmark timing.

[TPC-H.html.zip](https://github.com/ursacomputing/arrowbench/files/7171518/TPC-H.html.zip)

We have a few questions that we need to anser:

* should we keep the `read_parquet|feather()` along side `open_dataset()` or just use the dataset processing? I can leave both in and only use one as a default (though if we were to do that, I would change the way they are specified).

* what our defaults should be (and what should be run on every commit in conbench — which should be the same thing though they could technically be different). I propose: 
  * scale: 1, 10, (possibly 100 if we can get the ursa machines to generate that much data)
  * engine: arrow
  * format: native, parquet, feather (with the two files being driven by datasets)
  * mem_map: false (only applies to native, this will make the first iteration more in line with the others)

jonkeane:
@nealrichardson I'm planning to merge this today, unless you have any other things you would like to change or more time to look at it.

PR #37, opened 2021-09-23
Adding in more TPC-H queries

jonkeane:


R/bm-tpc-h.R
```diff
@@ -218,20 +232,397 @@ tpc_h_queries[[1]] <- function(input_func) {
...
+    select(l_extendedprice, l_discount, l_quantity, ps_supplycost, n_name, o_orderdate)
+
+  aggr <- all %>%
+    # kludge, should not need to collect here, but if we don't the answers we
+    # get are wrong (though they *do* complete).
+    collect() %>%
+    mutate(
+      nation = n_name,
+      # kludge, o_year = as.integer(format(o_orderdate, "%Y")),
+      o_year = as.integer(strftime(o_orderdate, "%Y")),
```
nealrichardson, line 574: perhaps, or why not?

```suggestion
      o_year = year(o_orderdate),
```

R/bm-tpc-h.R
```diff
@@ -218,20 +232,397 @@ tpc_h_queries[[1]] <- function(input_func) {
...
+
+  loc <- inner_join(lo_aggr, c, by = c("o_custkey" = "c_custkey"))
+
+  locn <- inner_join(loc, input_func("nation") %>% select(n_nationkey, n_name),
+                     by = c("c_nationkey" = "n_nationkey"))
+
+  res <- locn %>%
+    select(o_custkey, c_name, revenue, c_acctbal, n_name,
+           c_address, c_phone, c_comment) %>%
+    # kludge, should not need to collect here
```
nealrichardson, line 618: link to the jira where we'll fix this? also `compute()` instead?

R/bm-tpc-h.R
jonkeane: The only reason is I didn't realize we had all of those functions implemented already! I need to go back and run all of these through duckdb too. We don't *need* to keep these pipelines compatible with both if we don't want to, but it's a nice sanity check to be able to run it against a second backend (especially when we go to larger scales where we don't have the full answers)

jonkeane: duckdb might have `year()` or dbplyr might translate it appropriately without us needing to do anything, but that's one reason I haven't been simplifying these as I go

nealrichardson: got it, this is copied from the other repo

jonkeane: Yeah, and even when we do starting drafting them on their own keeping them compatible is worth _a bit_ of effort.

It also shouldn't be too hard (though I've not put in the time to do it) to run these against pure-R dplyr too. It should be "just" making a new set of `input_funcs`

R/bm-tpc-h.R
jonkeane: I added the link. I've also added to the other places we use this pattern to do `arrange() %>% compute() %>% head() %>% collect()` which is more characters, but should be a bit faster since we're not collecting everything into R

PR #43, opened 2021-10-04
Capture output from the benchamrks as they run and save it in an output field

### LINKED ISSUE #42: Retain the R output in the results json
Sometimes there might be important info in the results output, it would be nice if this were retained in the json output instead of throwing it away totally

jonkeane:
resolves #42


jonkeane:
@paleolimbot Would you mind looking over this PR for me?

PR #48, opened 2021-10-07
Add ability to run tpch with dplyr

jonkeane:


jonkeane:
incorporated into #55

Issue #50, opened 2021-11-02
TPC-H benchmarks 11-22


jonkeane:
[TPC-H spec](http://tpc.org/tpc_documents_current_versions/pdf/tpc-h_v3.0.0.pdf) The queries start on page 29. They are also included in a slightly more readable form at https://github.com/duckdb/duckdb/tree/master/extension/tpch/dbgen/queries

[Our current implementations of 1-10](https://github.com/ursacomputing/arrowbench/blob/main/R/bm-tpc-h.R#L219-L640)

paleolimbot:
Commenting so it shows up on my GitHub issue list!

jonkeane:
Here's an example command to run one of these (_most_ of the parameters have documentation in the readme, though we have a few that need more):

```
library(arrowbench)

withr::with_envvar(
  # ARROWBENCH_LOCAL_DIR is where to store the input data files (after generating
  # them) and the results, and any packages-to-be-benchmarked (see `lib_path` below)
  list(ARROWBENCH_LOCAL_DIR = "~/repos/ab_store/het"),
  {
    results <- run_benchmark(
      tpc_h,
      scale_factor = 1,
      query_id = 1,
      # cpu_count defaults to both 1 and max available, but feel free to set otherwise
      cpu_count = 8,
      format = c("native", "parquet"),
      # This is under-documented, but this string will actually install a whole
      # separate arrow package so it doesn't collide with your develeopment
      # version. Alternatively you can specify lib_path = "latest" which will
      # use whatever arrow version is installed in the .libPath your session is 
      # pointing to. 
      lib_path = "remote-apache/arrow@HEAD",
      engine = "arrow",
      n_iter = 1
    )
    
    all_results <- as.data.frame(results)
  }
)
```

PR #51, opened 2021-11-10
Add tpc-h 21 + 22

jonkeane:
A test to confirm if query 22 will work

PR #52, opened 2021-11-10
TCP-H queries 11-21

paleolimbot:
Just a work-in-progress PR as I get these updated! (Flagging #50)

### DIFF
--- a/R/bm-tpc-h.R
+++ b/R/bm-tpc-h.R
@@ -1175,14 +1175,14 @@ tpc_h_queries[[20]] <- function(input_func) {
       l_shipdate >= as.Date("1994-01-01"),
       l_shipdate < as.Date("1995-01-01")
     ) %>%
-    inner_join(partsupp_forest_ca, by = c("l_partkey" = "ps_partkey")) %>%
-    group_by(l_partkey, ps_suppkey) %>%
-    summarise(qty_threshold = 0.5 * sum(l_quantity), .groups = "drop")
+    semi_join(partsupp_forest_ca, by = c("l_partkey" = "ps_partkey", "l_suppkey" = "ps_suppkey")) %>%
+    group_by(l_suppkey) %>%
+    summarise(qty_threshold = 0.5 * sum(l_quantity))
 
   partsupp_forest_ca_filtered <- partsupp_forest_ca %>%
     inner_join(
       qty_threshold,
-      by = c("ps_suppkey", "ps_partkey" = "l_partkey")
+      by = c("ps_suppkey" = "l_suppkey")
     ) %>%
     filter(ps_availqty > qty_threshold)
 
--- a/R/bm-tpc-h.R
+++ b/R/bm-tpc-h.R
@@ -1163,26 +1163,23 @@ tpc_h_queries[[20]] <- function(input_func) {
     ) %>%
     select(s_suppkey, s_name, s_address)
 
-  partsupp_ca <- input_func("partsupp") %>%
-    semi_join(supplier_ca, c("ps_suppkey" = "s_suppkey"))
-
   part_forest <- input_func("part") %>%
     filter(grepl("^forest", p_name))
+
+  partsupp_forest_ca <- input_func("partsupp") %>%
+    semi_join(supplier_ca, c("ps_suppkey" = "s_suppkey")) %>%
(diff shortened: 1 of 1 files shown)


R/bm-tpc-h.R
```diff
@@ -639,6 +639,83 @@ tpc_h_queries[[10]] <- function(input_func) {
...
+  #   arrange(desc(value)) %>%
+  #   select(-global_value) %>%
+  #   collect()
+
+  # ...but not in Arrow yet, so we do a summarize() + left_join()
+  global_agr <- joined_filtered %>%
+    mutate(global_agr_key = 1L) %>%
+    group_by(global_agr_key) %>%
+    summarise(
+      global_value = sum(ps_supplycost * ps_availqty) * 0.0001000000
```
jonkeane, line 698: It's possible we're running into floating point precision issues here. The TPC-H spec is all-decimal, though our support for that from R is not super great. You might try `Scalar$create(0.0001000000)$cast(decimal(10, 10))`

```
format(0.0001000000, digits = 22)
[1] "0.0001000000000000000047922"
```

When I ran into this before, I ended up [using round](https://github.com/ursacomputing/arrowbench/blob/main/R/bm-tpc-h.R#L404-L407) though looking back, probably should have been more direct like this!

R/bm-tpc-h.R
```diff
@@ -639,6 +639,83 @@ tpc_h_queries[[10]] <- function(input_func) {
...
+    )
+
+  partkey_agr <- joined_filtered %>%
+    group_by(ps_partkey) %>%
+    summarise(value = sum(ps_supplycost * ps_availqty))
+
+  result2 <- joined_filtered %>%
+    mutate(global_agr_key = 1L) %>%
+    left_join(global_agr, by = "global_agr_key") %>%
```
jonkeane, line 707: This _should_ work. I took [a slightly different approach](https://github.com/ursacomputing/arrowbench/pull/51/files#diff-b47df652daa060d3d75dd06b6b88ea3772a7e83a16a15db1af70ce36017db401R678-R690) where I `collect()`ed the global aggregate and then pulled out the only element from it (I could and probably should do that in a slightly less fragile way). 

Something we might want to make a Jira for (though there might be one already or this might be implemented and I just didn't have the right call at my fingertips!) is if we can `pull()` a single value out of a query like that which would let us not `collect()` the global aggregate at first.

R/bm-tpc-h.R
```diff
@@ -639,6 +639,83 @@ tpc_h_queries[[10]] <- function(input_func) {
...
+    left_join(partkey_agr, by = "ps_partkey") %>%
+    filter(value > global_value) %>%
+    arrange(desc(value)) %>%
+    select(-global_agr_key, -global_value) %>%
```
jonkeane, line 711:
```suggestion
    select(ps_partkey, value) %>%
```

I've been using the positive selection like this since it's in the query itself. I agree that  `select(-global_agr_key, -global_value)` is more likely for what someone would do in R if they were writing this, but the queries themselves do have which columns will be selected, so I think we can go with that.

jonkeane: Also oddly(?) when I pulled this branch locally the `value` column didn't seem to show up in the error messages I got. I didn't dig in to why, but it's totally possible something is funky (in Arrow's hash join implementation) with the join / that name. Or possibly it's a bug in how we do negative selection? That might be worth investigating (separately from this, and can be deferred!)

jonkeane (changes requested):
This is taking shape, and is really great. Thanks! I've got a few comments / suggestions that might impact the validity check. Though I suspect the `select()` change alone might fix it.

jonkeane:
Thanks for this! I'll take a look at this later today. Do you mind if I push commits to your branch if I find anything?

jonkeane:
Wow, 20 was, indeed, a doozy. It took me a lot of staring and reasoning about that correlated subquery to figure out what the issue was, and the change ended up being minor: https://github.com/ursacomputing/arrowbench/pull/52/commits/87e7b8ede661df95d4142e18317115b739026027 though getting there took a lot.

jonkeane:
So far, I've tried to harmonize the format of `collect()`/saving outputs and a few other things (as well as removing kludges that aren't needed anymore), all of the queries work and validate, which is fantastic. I'll review them more deeply sometime today, and merge / enable them in conbench

PR #55, opened 2021-11-25
Add validation to scale factors 0.01-10

jonkeane:
This also has a bunch of clean up / useability enhancements:

* `tpch_answer()` function for retrieving an answer
* Move tpch queries to a separate file
* Caught a few bugs with extensions, filenames, and dots
* Caught a bug where changing the compression/format of a data source lost type information (e.g. reading a decimal in to R which then is saved as a float later)


jonkeane:
Aaah yes, query 12 doesn't work on any machine I've tried at scale factor 10 (so I don't have an answer there yet, I probably could + should craft an answer parquet for that so that it's there when the query engine is fixed to run it). I think that is a query engine problem (so will eventually be an apache/arrow jira) but I haven't gotten to the bottom of what the issue is, so haven't created one yet. Would you mind taking a look and seeing if you can get a reprex for that query and make a Jira for it?

jonkeane:
Oh fantastic, if this works that would be even better. You "just" moved lineitem to be the first table, did all the filtering there then joined after instead of starting with orders and then attempting to join to the filtered lineitem to that, yeah?

I'll use this new code here, but it's a bit funny that we get a crash in the first case — so we probably still want to dig into that and create a Jira for that (even if the fix is "just" to add a gate that stops someone from doing whatever is causing the crash in the first — I don't _think_ it's as simple as a right-hand table being filtered, cause we have other examples of that).

jonkeane:
Yeah, absolutely. Thanks again for that quick fix + review

PR #62, opened 2021-12-09
Use a single query for tpch-22

### LINKED ISSUE #61: Clean up query 22 to have a single query
Currently we `collect()` and then use it in the next query, instead we should make a trivial join column and then join it as a subquery.

jonkeane:
Resolves #61


jonkeane:
@paleolimbot Would you mind taking a look at this?

paleolimbot (approved):
This looks good to me!

PR #64, opened 2021-12-20
Let me use DuckDB with sql

jonkeane:
This adds functionality to query SQL backends with the SQL-based TPC-H queries. Specifically, it adds a "duckdb_sql" engine that can be driven by parquet files or by native in-memory tables.

Left for future work is factoring out the input setup to make it easier to run one-off TPC-H queries.


jonkeane:
@paleolimbot Would you mind taking a look at this when you get a chance?

R/bm-tpc-h.R
```diff
@@ -114,13 +114,26 @@ tpc_h <- Benchmark("tpc_h",
...
+          sql_query <- paste0("CREATE TABLE IF NOT EXISTS ", name, " AS SELECT * FROM parquet_scan('", file, "');")
         }
 
         DBI::dbExecute(con, sql_query)
       }
+
+      # If we are using the SQL engine, then we need to replace tpc_h_queries
+      if (engine == "duckdb_sql") {
+        tpc_h_queries <- list()
+        # TODO: should this actually just be query_id?
```
paleolimbot, line 128: Maybe solve this instead of leaving it as a note? I think that you have `query_id` as a variable here already and so it seems like you only need one query and not all of them?

R/bm-tpc-h.R
```diff
@@ -157,14 +170,18 @@ tpc_h <- Benchmark("tpc_h",
...
   },
   # after each iteration, check the dimensions and delete the results
   after_each = {
     # If the scale_factor is < 1, duckdb has the answer
     if (scale_factor %in% c(0.01, 0.1, 1, 10)) {
       answer <- tpch_answer(scale_factor, query_id)
 
+      # the result is sometimes a data.frame, turn into a tibble for printing
+      # purposes
+      result <- dplyr::as_tibble(result)
```
paleolimbot, line 183: I was wondering why some of them printed pretty and some didn't!

R/ensure-lib.R
```diff
@@ -261,13 +261,12 @@ ensure_custom_duckdb <- function() {
   # We need to check if the installed duckdb has the tpch extension built. If it
   # does not, we will build it (with the appropriate envvars to build with tpch)
 
-  # We check if duckdb can generate a (very small) set of tpch data to ensure it
-  # has the tpch extension. This is done in a call to `system` so that we don't
-  # load the duckdb namespace/dll before installing it. In my testing even pkgload::unload()
-  # couldn't fully unload duckdb.
+  # We check if duckdb has the tpch extension. This is done in a call to `system`
+  # so that we don't load the duckdb namespace/dll before installing it. In my
+  # testing even pkgload::unload() couldn't fully unload duckdb.
```
paleolimbot, line 266: Again, battle for another day, but you could install the custom duckdb into a separate library path in the local cache directory and use something like `callr:r()` (which I think has an arg for library path) to do all the custom duckdb stuff. The 'pak' package does something like this to keep its own dependencies from interfering with the packages that it's installing.

paleolimbot (commented):
Just a few notes, not all are relevant to this PR but things to think about.

I think you should avoid passing `con` down into the query functions...there are a few ways you could go about avoiding this, one being downloading the queries on demand and/or caching them; the other being to just create a duckdb connection every time which I don't think is expensive for fetching the queries.

I put a note in about sandboxing your duckdb environment, which would mean calling out to a subprocess everytime you need the custom version. Not necessary here, but might help for those of us who regularly install the normal duckdb and have to recompile it whenever we run arrowbench.

R/bm-tpc-h.R
```diff
@@ -341,3 +358,41 @@ tpch_answer <- function(scale_factor, query_id, source = c("arrowbench", "duckdb
...
+  # wrap the SQL in a function
+  function(input_func = NULL, collect_func = NULL, con) {
+    DBI::dbGetQuery(con, query_sql)
+  }
+}
+
+
+get_sql_tpch_query <- function(con, query_num) {
+  ensure_custom_duckdb()
+
```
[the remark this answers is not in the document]
jonkeane: _nods_ agreed on the `con` front. I'm re-working the code to do like you suggest and create a new one for each. 

I'm a bit hesitant to grab the queries from github since that's fragile both with the location in their git repo, github being up, and needing an active internet connection (though the current install process relies on that as well, there should soon be a way to avoid that)

R/bm-tpc-h.R
jonkeane: I think that's actually a separate issue — the data.frames here are all coming from `DBI::dbGetQuery(...)`. 

I started to (then put down in the interest of time) changing all of these to waldo which is so so so much better for printing this info. I've made #66 to do that.

R/ensure-lib.R
jonkeane: Yeah, I like this idea a lot (though think I will defer that for another day). We actually already have some infrastructure for creating custom library paths and installing into on them in arrowbench, we could re-use some of that (especially with a `callr::r()` call or the like)

R/bm-tpc-h.R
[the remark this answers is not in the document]
paleolimbot: Over the break I was thinking...could you use package data for this? So you could do `arrowbench::tcph_queries[12]`? That way you could have a script that updates the queries from GitHub but they'd be static until you run the update script again.

R/ensure-lib.R
paleolimbot: When I'm back I'll PR this in if I get a chance!

jonkeane: Cool, I've written https://github.com/ursacomputing/arrowbench/issues/67 along with a few links to the extension install work in DuckDB that will make this whole thing much easier (we will be able to install DuckDB from CRAN and then install the extension at runtime)

R/bm-tpc-h.R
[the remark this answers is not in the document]
jonkeane: Oh, I like that — though we could even pull them from duckdb directly in that script (it's a bit less of a concern of suddenly breaking / needing internet access, but I suspect that the duckdb tpch query interface is slightly more stable than the location of the files). Maybe we could add that when we look at #67 ?

Issue #67, opened 2021-12-30
Call duckdb custom calls in a subprocess (with a unique lib path)


jonkeane:
So that one doesn't need to constantly need to rebuild duckdb switching between released and with-tpch-enabled versions:

See https://github.com/ursacomputing/arrowbench/pull/64#discussion_r772743721

This will be _slightly_ less annoying once DuckDB releases a new release now that [DuckDB can install extensions at runtime](https://github.com/duckdb/duckdb/pull/2569)

PR #68, opened 2022-01-04
Install custom duckdb in a separate package library and use it!

paleolimbot:
A stab at #67!


R/custom-duckdb.R
```diff
@@ -0,0 +1,130 @@
...
+      } else {
+        rlang::abort(
+          "An unexpected error occured whilst querying TPC-H enabled duckdb",
+          parent = e
+        )
+      }
+    }
+  )
+
+  if (identical(result$query_nr, 1L)) {
```
jonkeane, line 27:
```suggestion
  # Check that the result has a query in it
  if (identical(result$query_nr, 1L)) {
```

Or some other kind of classificatory statement?

tests/testthat/test-bm-tpc-h.R
```diff
@@ -83,13 +88,25 @@ ORDER BY
...
 
   # clean up
   DBI::dbDisconnect(con_one, shutdown = TRUE)
-})
\ No newline at end of file
+})
+
+test_that("tpch_answer() works", {
+
+})
```
jonkeane, line 112: It looks like this can be removed? You have the test up above that does both of the options here, yeah?

tests/testthat/test-bm-tpc-h.R
```diff
@@ -83,13 +88,25 @@ ORDER BY
...
+    engine = "duckdb",
+    scale_factor = 0.001,
+    format = "parquet",
+    query_id = 1,
+    con = con_one
+  )
 
   query_01_func <- get_sql_query_func(1)
-  query_01_func(con = con_one)
+  expect_s3_class(query_01_func(con = con_one), "data.frame")
```
jonkeane, line 104: Nice addition — this is a bit of scope creep, but I wonder if it would be possible to also test it against our answers (though we would then run the risk that duckdb breaks and doesn't actually work for this one query / at this sacle factor but that's not our fault...)

R/custom-duckdb.R
```diff
@@ -0,0 +1,130 @@
...
+    con <- DBI::dbConnect(duckdb::duckdb(dbdir = dbdir))
+    on.exit(DBI::dbDisconnect(con, shutdown = TRUE))
+    res <- DBI::dbSendQuery(con, sql, arrow = TRUE)
+
+    # this could be streamed in the future when the parquet writer
+    # in R supports streaming
+    reader <- duckdb::duckdb_fetch_record_batch(res)
+    table <- reader$read_table()
+    arrow::write_parquet(table, sink)
+    sink
```
jonkeane, line 90: Am I reading the callr docs / code right that the way that callr gets the data from the subprocess to the main one is by saving and RDS, so this avoids attempting to serialize a large table by saving it with arrow in the first place?

jonkeane (commented):
This is looking really good — thank you for sending it. A few questions /  comments to make sure I'm fully following this.

tests/testthat/test-bm-tpc-h.R
paleolimbot: When I do this I get:

```r
expect_identical(query_01_func(con = con_one), tpch_answer(0.001, 1))
#>  Error in tpch_answer(0.001, 1) : 
#>  The answer file (looking for arrowbench/tpch/answers/scale-factor-0.001/tpch-q01-sf0.001.parquet in the arrowbench package directory) was not found 
```

paleolimbot: Maybe there's another way that I didn't consider?

R/custom-duckdb.R
paleolimbot: Yeah! That's the idea. The streaming might make it possible to get bigger scales, too (I'll check to see I can make this work already by bypassing `read_parquet()`).

paleolimbot: (I did try a workaround but it's too verbose for here, I think: https://gist.github.com/paleolimbot/bab9cb895f90f64d21968d48b3ef9c1f#gistcomment-4036437 )

tests/testthat/test-bm-tpc-h.R
jonkeane: Oh, no that's totally right — I don't have answers saved for 0.001 just 0.01, 0.1, 1, and 10. Let's leave this here — we haven't had any issues with this (and we test these when we create them in the answer creation script too, so we should see it then if we run into it...)

R/custom-duckdb.R
jonkeane: Yeah, I think we should wait for those arrow tickets to get done to do this right here instead of trying to hack it in (and a good impetus to push on those tickets!). We _soon_ will be able to at least stream from DuckDB when they release their next release, so that'll improve things a bit.

PR #71, opened 2022-01-20
Allow one to change the compression of files for TPC-H

jonkeane:
This adds a passing of the `compression` argument to `get_input_func()` so it's easier to test different compressions when using the pattern:

```
library(arrowbench)
library(dplyr, warn.conflicts = FALSE)
library(arrow, warn.conflicts = FALSE)

input_funcs <- get_input_func(
        engine = "arrow",
        scale_factor = 0.01,
        query_id = 1,
        format = "parquet"
      )

get_query_func(1, "arrow")(input_funcs)
```

### DIFF
--- a/R/bm-tpc-h.R
+++ b/R/bm-tpc-h.R
@@ -203,14 +203,16 @@ find_input_func <- function(func) {
 #' @param scale_factor what scale factor to reference
 #' @param query_id which query is being used
 #' @param format which format
+#' @param compression which compression to use (default: "uncompressed")
 #' @param con a connection
-#' @param memory_map should the file be memory mapped?
+#' @param memory_map should the file be memory mapped? (only relevant for the "native" format with Arrow)
 #'
 #' @export
 get_input_func <- function(engine,
                            scale_factor,
                            query_id,
                            format,
+                           compression = "uncompressed",
                            con = NULL,
                            memory_map = FALSE) {
   # ensure that we have the _base_ tpc-h files (in parquet)
@@ -232,7 +234,8 @@ get_input_func <- function(engine,
       tpch_files[tpch_tables_needed],
       ensure_format,
       FUN.VALUE = character(1),
-      format = format_to_convert
+      format = format_to_convert,
+      compression = compression
     )
 
     # specify readers for each format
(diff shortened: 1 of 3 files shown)


R/bm-tpc-h.R
```diff
@@ -203,14 +203,16 @@ find_input_func <- function(func) {
 #' @param scale_factor what scale factor to reference
 #' @param query_id which query is being used
 #' @param format which format
+#' @param compression which compression to use (default: "uncompressed")
 #' @param con a connection
-#' @param memory_map should the file be memory mapped?
+#' @param memory_map should the file be memory mapped? (only relevant for the "native" format with Arrow)
```
westonpace, line 208: This confuses me.  My understanding of the "native" format is that a table is passed into the exec plan.  So why would memory mapping have any role here?

jonkeane: Yeeeah, these names are probably more confusing than they need to be. FWIW, the default for `memory_map` is the one we want for arrowbench. But here it goes: 

"native" format uses `read_feather(file, as_data_frame = FALSE)` to get an Arrow table. However, by default that will use memory mapping so the first time it's called `read_feather(file, as_data_frame = FALSE)` is basically instantaneous since it's not really reading the data in, but then the first query run will be the time it takes to copy the data from disk to memory and then the duration of the query. But because of caching, the second run will be just the query. So for native formats we disable memory mapping in order to ensure we have the full table read in (_before_ we start measuring the query run time) first.

jonkeane: And for our file-based formats "parquet" and "feather", we're using `open_dataset()` which preforms better since even with a single file, we can do some filtering/column reading pushdown with that as opposed to `read_parquet(...) %>% the rest of the query`

westonpace: That makes sense now.  Also, even without column/filter pushdown `open_dataset` will allow you to start processing the data before the entire table is read in (assuming the file has more than one row group of course :laughing: )

westonpace: Although I think I'd lean towards just not having `mmap` as a parameter (and always specifying false).

jonkeane: That makes sense, yeah maybe I will pull it out at some point. But to keep this PR small I'm going to leave it in + merge

PR #72, opened 2022-01-21
Adds the ability to specify the number of row groups when creating parquet/arrow files

westonpace:


jonkeane:
Thanks for this, this is amazing! I have a few thoughts that I'll push to this branch if you don't mind

R/ensure-format.R
```diff
@@ -90,26 +94,34 @@ ensure_format <- function(
...
+#' @param num_groups how many rows groups to create (if the format supports it)
 #'
 #' @return the write function to use
 #' @export
-get_write_function <- function(format, compression) {
+get_write_function <- function(format, compression, num_groups) {
   force(compression)
   if (format == "feather") {
-    return(function(...) arrow::write_feather(..., compression = compression))
+    return(function(x, ...) arrow::write_feather(x, ..., chunk_size = get_chunk_size(x, num_groups) %||% 65536L, compression = compression))
```
jonkeane, line 122: `%||% 65536L` is a bit of magic that will return `65536L` if the thing before the `%||%` is `NULL`. That number is the default for `write_feather` and we (unfortunately) can't pass `NULL` to mean default there like we can with parquet.

westonpace:
You are absolutely right wrt chunk_size.  I've converted this PR from num_groups to chunk_size.

PR #78, opened 2022-04-11
Add csv writing benchmark. Closes #76

### LINKED ISSUE #76: Add a csv writing benchmark
Similar to https://github.com/ursacomputing/arrowbench/blob/main/R/bm-read-csv.R but writing instead of reading.

boshek:


R/bm-write-csv.R
```diff
@@ -0,0 +1,74 @@
...
+  },
+  # the benchmark to run
+  run = {
+    write_csv_func(df, result_file)
+  },
+  # after each iteration, check the dimensions and delete the results
+  after_each = {
+    stopifnot(file.exists(result_file))
+    unlink(result_file)
+  },
```
jonkeane, line 41: Checking that the file exists is good, I wonder if there's something else we could do to confirm that this has worked correctly?

R/bm-write-csv.R
```diff
@@ -0,0 +1,74 @@
...
+#' @section Parameters:
+#' * `source` A CSV file path to write to
+#' * `writer` One of `c("arrow", "data.table", "vroom", "readr",)`
+#' * `input` One of `c("arrow_table", "data_frame")`
+#'
+#' @export
+write_csv <- Benchmark(
+  "write_csv",
+  setup = function(source = names(known_sources),
+                   writer = c("arrow", "data.table", "vroom", "readr", "utils"),
```
jonkeane, line 12:
```suggestion
                   writer = "arrow",
```

This is super subtle (and not super well documented) but what's listed here will be what's run by default if this isn't otherwise specified. So using only `"arrow"` will mark that as the default, then we add the others to `match.arg` below

R/bm-write-csv.R
```diff
@@ -0,0 +1,74 @@
...
+                   input = c("arrow_table", "data_frame")) {
+    writer <- match.arg(writer)
```
jonkeane, line 14:
```suggestion
    writer <- match.arg(writer, c("arrow", "data.table", "vroom", "readr", "utils"))
```

R/bm-write-csv.R
```diff
@@ -0,0 +1,74 @@
...
+    input <- match.arg(input)
+
+    # source defaults are retrieved from the function definition (all available
+    # known_sources) and then read the source in as a data.frame
+    source <- ensure_source(source)
+    df <- read_source(source, as_data_frame = match.arg(input) == "data_frame")
+
+    # Map string param name to functions
+  BenchEnvironment(
```
jonkeane, line 23:
```suggestion
    BenchEnvironment(
```

Super minor style thing

R/bm-write-csv.R
```diff
@@ -0,0 +1,74 @@
...
+  before_each = {
+    result_file <- tempfile(fileext = ".csv")
+  },
+  # the benchmark to run
+  run = {
+    write_csv_func(df, result_file)
+  },
+  # after each iteration, check the dimensions and delete the results
+  after_each = {
+    stopifnot(file.exists(result_file))
```
jonkeane, line 39:
```suggestion
    stopifnot("Output file does not exist" = file.exists(result_file))
```

This is relatively new (and a bit extra), but helps make clear what happened to casue the stop.

R/bm-write-csv.R
```diff
@@ -0,0 +1,74 @@
...
+    unlink(result_file)
+  },
+  valid_params = function(params) {
+    ## Only arrow fns will accept an arrow_table
+    drop <- ( params$input == "arrow_table" & params$writer != "arrow" )
+    params[!drop,]
+  },
+  packages_used = function(params) {
+    params$writer
```
jonkeane, line 48: We might want to strip out `"utils"` here. I suspect it'll work fine, but we'll get the warning `package ‘utils’ is a base package, and should not be updated`

jonkeane: Separately from that, I wonder if we should call this `"base"`. `"utils"` is most accurate, but I suspect most folks would call it `"base"`. I'm not the fence, what do you think?

jonkeane (changes requested):
Thanks for this, this is really great. A few comments + suggestions. 

Most of the benchmarks we've written so far have _not_ been tested (we thought it was easier to rely on benchmarking itself to test them), but I've been thinking we might want to add tests (at least | especially for new benchmarks). What do you think about adding a integration-style test for this one (basically run it with one of the tiny sample files we have and make sure it doesn't error with good inputs)?

R/bm-write-csv.R
boshek: I actually _did_ read that but it didn't sink in until right now. Thanks.

R/bm-write-csv.R
```diff
@@ -0,0 +1,74 @@
...
+#' * `source` A CSV file path to write to
+#' * `writer` One of `c("arrow", "data.table", "vroom", "readr",)`
+#' * `input` One of `c("arrow_table", "data_frame")`
+#'
+#' @export
+write_csv <- Benchmark(
+  "write_csv",
+  setup = function(source = names(known_sources),
+                   writer = c("arrow", "data.table", "vroom", "readr", "utils"),
+                   input = c("arrow_table", "data_frame")) {
```
jonkeane, line 13: Oh and one more thing: should we add `compression` here too? at least arrow can write compressed csvs. I believe most (all?) of the others can too, though some of the docs don't make that super obvious

R/bm-write-csv.R
boshek: Yeah actually for intent I think `"base"` is probably best. I was sort of on auto-pilot which led me to `"utils"`.

R/bm-write-csv.R
boshek: What about checking the dimensions like this?
```r
after_each = {
    stopifnot(identical(dim(df), dim(arrow::open_dataset(result_file, format = "csv"))))
    stopifnot("Output file does not exist" = file.exists(result_file))
    unlink(result_file)
  }
```

boshek:
@jonkeane I think this is ready for a re-review

tests/testthat/test-bm-write-csv.R
```diff
@@ -0,0 +1,12 @@
...
+    run_benchmark(
+      write_csv,
+      source = "nyctaxi_sample",
+      writer = c("arrow", "data.table", "vroom", "readr", "base")
+      ), "arrowbench_results"
+  )
+})
+
+
+
```
jonkeane, line 12:
```suggestion
})

```

Super minor nitpick that lintr would complain about if we had it on here

jonkeane (approved):
Looks good, thank you!

PR #79, opened 2022-04-11
[MINOR] some package fixes

boshek:
Just a couple things I noticed.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -5,4 +5,5 @@
 tests/testthat/results
 .DS_Store
 results/
+data/
 .vscode/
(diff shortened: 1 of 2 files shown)


.gitignore
```diff
@@ -5,4 +5,5 @@
 tests/testthat/results
 .DS_Store
 results/
+data/
```
jonkeane, line 8: We might also want to add `tests/testthat/data/` here too

boshek: good call. done.

jonkeane (approved):
Looks good! Thank you

PR #82, opened 2022-04-18
[ENG-3630] Use `waldo::compare()` for validating TPC-H queries

alistaire47:
Changes from `all.equal()` to [`waldo::compare()`](https://waldo.r-lib.org/reference/compare.html) in TPC-H benchmark for nicer diffs and adjusts printing and control flow (`compare()` returns an empty vector instead of `TRUE` when there are no differences) accordingly. Leaves programmatic usages of `all.equal()` (those not printed) alone. 

Minor changes:

- Documents `ARROWBENCH_TEST_CUSTOM_DUCKDB` in README
- Specifies `repos` in `install.packages()` call in tests


jonkeane:
This is really great! Did you have a chance to try also swapping out https://github.com/ursacomputing/arrowbench/blob/7882f37c39bf7a9918f537257fd0db41ba121c88/R/bm-tpc-h.R#L129 as well? it looks like it worked for line 106 up above.

alistaire47:
> This is really great! Did you have a chance to try also swapping out
> 
> https://github.com/ursacomputing/arrowbench/blob/7882f37c39bf7a9918f537257fd0db41ba121c88/R/bm-tpc-h.R#L129
> as well? it looks like it worked for line 106 up above.

I did, but it gets grepped below

https://github.com/ursacomputing/arrowbench/blob/7882f37c39bf7a9918f537257fd0db41ba121c88/R/bm-tpc-h.R#L133-L139

and waldo changes those messages, specifically to

``` r
> dput(waldo::compare(Sys.Date(), as.character(Sys.Date()), ignore_attr = FALSE, tolerance = 0.01))
structure("`old` is \033[32man S3 object of class <Date>, a double vector\033[39m\n`new` is \033[32ma character vector\033[39m ('2022-04-18')", class = "waldo_compare", max_diffs = 10)
```


Since that one doesn't get printed (the variable gets overwritten on 139), I figured there was less benefit, and a little risk that a more complicated regex might catch something inadvertently. If you're ok with a more complicated regex we can swap it, though.

jonkeane:
Ah, yup yup you're right on that! Thanks

R/bm-tpc-h.R
```diff
@@ -103,12 +103,10 @@ tpc_h <- Benchmark("tpc_h",
       # of the query validation output data when rounded to the nearest 1/100th
       # > For results from SUM aggregates, the resulting values must be within
       # $100 of the query validation output data.
-      all_equal_out <- all.equal(result, answer, tolerance = 0.01)
+      all_equal_out <- waldo::compare(result, answer, tolerance = 0.01)
```
jonkeane, line 106: `waldo::compare` is totally fine here, but we have check note complaining about not importing anything from {waldo}. We do use `compare` again below, so it's probably easiest to actually import compare with a roxygen tag + get this Note out of the way at the same time 

```
❯ checking dependencies in R code ... NOTE
  Namespace in Imports field not imported from: ‘waldo’
    All declared Imports should be used.
```

jonkeane (changes requested):
Thanks for this, this is very very close. One small change about importing from `waldo` to avoid a Note.

Also, to confirm: You ran the tests locally as well to make sure the change worked, yeah? We also could enable these tests in CI, though that will make our CI runs much much longer just to install duckdb, so I'm hesitant to do that.

R/bm-tpc-h.R
alistaire47: Huh, that seems like an odd choice in r-devel; `::`-only imports are everywhere. Added a roxygen import but didn't remove the namespacing so it's still easy to see where it comes from.

And yes, I ran tests with the DuckDB tests enabled and it passes; just lots of those deparse warnings.

jonkeane: Awesome, thanks!

jonkeane (approved):
Looks good, thanks for this!

Issue #74, opened 2022-03-01
New benchmark for different record batch sizes


jonkeane:
There are a few places we should do this:

* reads from disk using Datasets (save parquets with varying row_group sizes, and see what the performance is. Prior art and hypotheses: https://lists.apache.org/thread/xt9z12bjkm7czvv38q993czd80bdgrlc
* Across the C-Data stream (to answer questions like: https://github.com/duckdb/duckdb/issues/3138)

alistaire47:
Closed via #87

PR #87, opened 2022-05-03
[ENG-3633] New benchmark for different record batch sizes

alistaire47:
Adds a new benchmark per #74 that tests querying data saved with different row group sizes with a pile of filters that will require scanning some row groups but not others. Currently only implements Fannie Mae, the biggest source, as the effect is small unless the data is big. Tried other params like compression to see if there was an interaction, but the only notable interaction I found is parallelism.


R/bm-row-group-size.R
```diff
@@ -0,0 +1,68 @@
...
+#' @section Parameters:
+#' * `source` A known-file id, or a file path to read in
+#' * `chunk_size` Number of rows to write in each row group. Suggested sizes:
+#'     `chunk_size = list(NULL, 10000L, 100000L, 1000000L)`
+#'
+#' @export
+row_group_size <- Benchmark(
+  "row_group_size",
+  setup = function(source = "fanniemae_2016Q4",  # TODO implement more sources
+                   chunk_size = NULL) {
```
boshek, line 12: What the reason that this is being left as `NULL`?

boshek (approved):

R/bm-row-group-size.R
alistaire47: That's the default for `write_parquet()`. `match.arg()` only works for character vectors, so there's not a great way to set a list of options/defaults here afaik; it's something to be aware of when using `default_params()` with this benchmark. We could write a more flexible version of `match.arg()` to make `default_params()` better, but it would confuse anybody new to this code. Or maybe we could store non-character defaults as attributes or something?

tests/testthat/test-bm-row-group-size.R
```diff
@@ -0,0 +1,12 @@
+test_that("row_group_size benchmark runs", {
+
+  params <- default_params(row_group_size, chunk_size = list(NULL, 10000L, 100000L, 1000000L))
+
+  expect_s3_class(
+    run_benchmark(
+      row_group_size,
+      chunk_size = list(NULL, 100000L)
+    ),
```
jonkeane, line 9: Is it possible to use the test dataset here? https://github.com/ursacomputing/arrowbench/blob/6ff8af0dd62d36a57fc7b9571753af360d7c9cb9/R/known-sources.R#L62-L66


The validation might need to be tweaked slightly (or maybe we pass a parameter to disable the dimension validation when we're testing?)

R/bm-row-group-size.R
```diff
@@ -0,0 +1,68 @@
...
+          | f21 > 10000
+          | f22 > 3e5
+          | f23 > 1e5
+          | f25 > 1000
+          | f26 > 5e4
+        ) %>%
+        collect()
+
+      result_dim <- c(514L, 31L)
+    }
```
jonkeane, line 56: We can do this as a follow on, but it would be nice to have a few different queries: one where we read the whole thing or select almost all of it, then ones like this one where we are pretty selective...

Would you mind making a follow on that's basically "expand this test to use different queries too"?

R/bm-row-group-size.R
```diff
@@ -0,0 +1,68 @@
...
+  },
+  # after each iteration, check the dimensions and delete the results
+  after_each = {
+    stopifnot(identical(dim(result), result_dim))
```
jonkeane, line 60: We can also use the named argument of `stopifnot()` to make this error slightly more helpful:

https://github.com/ursacomputing/arrowbench/blob/6ff8af0dd62d36a57fc7b9571753af360d7c9cb9/R/bm-remote-dataset.R#L24

(it's relatively new, so old Rs won't support it, but that's fine for this IMO

jonkeane (approved):
Looks good — a few small suggestions, most of which could be deferred to a future ticket(s) if we wanted to.

R/bm-row-group-size.R
alistaire47: added!

tests/testthat/test-bm-row-group-size.R
alistaire47: Ok so I'm not sure why this is happening, but when I tried to add this in, the names for `fanniemae_2016Q4` and `fanniemae_sample` are not the same; the former is `f0 ... f30` and the latter is `0 ... 30`. Also the dims listed there are totally wrong somehow; it should be `c(1000L, 31L)`.

I can change the dims easily, but I suspect the names are getting created automatically somewhere; I'm not sure where to fix this.

alistaire47: maybe I can just put `col_names = paste0('f', 0:31)` there? but that's inexplicably different than what the full source has

jonkeane: Oh weird. Those two data sets did come from a slightly different source. Let's keep the test as is now, but mark a todo to resolve it once we resolve the ticket I'm about to describe:

Could you make a ticket either here or in ursacomputing/benchmarks to harmonize these two test datasets?

alistaire47: Sure. But I realized I'm somehow wrong above; now the dim and names are right somehow and I don't understand what's happening...looking into it

alistaire47: ahhhhh, it's got piles of null columns. Still something went strange before; it looked totally different when I created it with `ensure_format()`

alistaire47: this is...weird

``` r
> read_parquet(input_file)
# A tibble: 757 × 108
      f0       f1     f2 f3    f4    f5       f6    f7    f8    f9   f10    f11   f12   f13    f14
   <???>  <int64>  <int> <chr> <chr> <chr> <???> <dbl> <dbl> <dbl> <???>  <dbl> <int> <int>  <int>
 1     .     1e11  82009 R     Other Other     .  5.38  5.38 55000     .     0    240 82009 102009
 2     .     1e11  92009 R     Other Other     .  5.38  5.38 55000     .     0    240 82009 102009
 3     .     1e11 102009 R     Other Other     .  5.38  5.38 55000     .     0    240 82009 102009
 4     .     1e11 112009 R     Other Other     .  5.38  5.38 55000     .     0    240 82009 102009
 5     .     1e11 122009 R     Other Other     .  5.38  5.38 55000     .     0    240 82009 102009
 6     .     1e11  12010 R     Other Other     .  5.38  5.38 55000     .     0    240 82009 102009
 7     .     1e11  22010 R     Other Other     .  5.38  5.38 55000     . 54351.   240 82009 102009
 8     .     1e11  32010 R     Other Other     .  5.38  5.38 55000     . 54219.   240 82009 102009
 9     .     1e11  42010 R     Other Other     .  5.38  5.38 55000     . 53955.   240 82009 102009
10     .     1e11  52010 R     Other Other     .  5.38  5.38 55000     . 53955.   240 82009 102009
# … with 747 more rows, and 93 more variables: f15 <int>, f16 <int>, f17 <int>, f18 <int>,
#   f19 <int>, f20 <int>, f21 <int>, f22 <int>, f23 <int>, f24 <int>, f25 <chr>, f26 <chr>,
#   f27 <chr>, f28 <int>, f29 <chr>, f30 <chr>, f31 <int>, f32 <int>, f33 <dbl>, f34 <chr>,
#   f35 <chr>, f36 <chr>, f37 <???>, f38 <???>, f39 <chr>, f40 <dbl>, f41 <chr>, f42 <???>,
#   f43 <int>, f44 <int>, f45 <dbl>, f46 <???>, f47 <???>, f48 <dbl>, f49 <???>, f50 <int>,
#   f51 <int>, f52 <int>, f53 <dbl>, f54 <dbl>, f55 <dbl>, f56 <dbl>, f57 <dbl>, f58 <dbl>,
#   f59 <???>, f60 <???>, f61 <???>, f62 <dbl>, f63 <dbl>, f64 <???>, f65 <???>, f66 <???>, …
> str(read_parquet(input_file)[[1]])
 'vctrs_unspecified' logi [1:757] NA NA NA NA NA NA ...
```

alistaire47: harmonizing arrowbench story: [issue](https://github.com/ursacomputing/arrowbench/issues/88), [story](https://voltrondata.clickup.com/t/14188074/ENG-4499)

jonkeane: Thanks! Yeah, I'm not totally surprised that there's some oddities in here. It was a bit bolted on | copied over in a bit of a hurry

alistaire47: Ok, I enabled `fanniemae_sample` for testing, but only for querying everything, not the filters, because all the columns are different for now. Put in a `TODO` referencing the story above; it will not unlikely get resolved as part of that story (which will likely break the full filters because of column name changes). We could make a separate story for it, but it's pretty trivial once the data is fixed.

R/bm-row-group-size.R
alistaire47: Went ahead and added this because it's pretty straightforward, and this works whereas filters are borked for `fanniemae_sample`. Pushing in a sec once all the tests pass locally

alistaire47: Ok some other tests failed locally (I think unrelated? Still debugging) so I went ahead and pushed so I can see what happens in CI. If it passes and you're good with the changes, we can merge

alistaire47: Yeah, all the failures are due to setting env vars so as not to spew files in the repo; testthat doesn't respect those env vars and runs everything relative to `test_path()`

PR #89, opened 2022-05-06
[ENG-3634] Add benchmarks that materialize altrep-backed vectors

### LINKED ISSUE #38: Add benchmarks that materialize altrep-backed vectors
We want to benchmark how quickly we can select out (e.g. `array[1:10]`) of and materialize the full altrep vector (i.e. `array[]`). We should write benchmarks that do both of these so we can measure them.

Additionally, we should confirm that the more macro benchmarks that can are taking advantage of altrep to see how it helps real workflows.

alistaire47:
Closes #38. Adds a benchmark that times how long it takes to materialize a subset or entire vector, with/without altrep enabled. Approach based on [this comment](https://github.com/apache/arrow/pull/11225#issuecomment-928981161), but lets NAs be handled by data sources (the `type_{type}` sources already have columns with varying proportions of NAs). Subsets data sources down to columns that can be stored as altrep.


R/bm-array-altrep-materialization.R
```diff
@@ -0,0 +1,119 @@
...
+
+  before_each = {
+    result <- NULL
+  },
+
+  run = {
+    result <- lapply(arrays, function(a) {
+      list(
+        array_subset = a[1:10],
+        array_materialized = a[]
```
boshek, line 93: I think a code comment here explaining why subsetting and return the full array is required. As you mentioned, linking here is probably a good idea at least: https://github.com/apache/arrow/pull/11225#issuecomment-928981161

R/bm-array-altrep-materialization.R
```diff
@@ -0,0 +1,119 @@
...
+  },
+
+  before_each = {
+    result <- NULL
+  },
+
+  run = {
+    result <- lapply(arrays, function(a) {
+      list(
+        array_subset = a[1:10],
```
boshek, line 92: Would it ever make sense to expose the `10` as an argument (and the one below) so we could benchmark the size of the subset?  
```suggestion
        array_subset = a[1:10],
```

boshek (approved):
Really awesome work. Some minor style things and a question but the approach seems sound to me.

R/bm-array-altrep-materialization.R
```diff
@@ -0,0 +1,119 @@
...
+    )
+    source <- match.arg(source, names(known_sources))
+
+    options(arrow.use_altrep = altrep)
+    path <- ensure_format(source, format = "parquet", compression = "snappy")
+
+    # exclude non-altrep types
+    pq <- arrow::ParquetFileReader$create(path)
+    table_arrow_types <- lapply(pq$GetSchema()$fields, function(x) x$type)
+    arrow_altrep_types = list(
```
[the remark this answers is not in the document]
alistaire47: Thanks! We should maybe lint...

R/bm-array-altrep-materialization.R
```diff
@@ -0,0 +1,119 @@
...
+#' @importFrom purrr transpose
+#' @export
+array_altrep_materialization <- Benchmark(
+  "array_altrep_materialization",
+
+  setup = function(
+      source = names(known_sources),
+      exclude_nulls = FALSE,
+      altrep = TRUE
+  ) {
```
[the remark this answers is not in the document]
jonkeane: Should we add stylr hooks to autostyle things like this? I suspect lintr will complain about _a lot_ of arrowbench because of the way it needs to do what it does, but styler should work ok + help make these easy.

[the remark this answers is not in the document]
alistaire47: As long as it doesn't break anything I'm in. The code isn't in too bad of shape that way though (imo), so it doesn't seem urgent.

R/bm-array-altrep-materialization.R
```diff
@@ -0,0 +1,123 @@
...
+    materialized_altrep_classes <- lapply(result_t[["array_materialized"]], function(x) .Internal(altrep_class(x)))
+    subset_is_altrep <- vapply(subset_altrep_classes, is.null, logical(1))
+    materialized_is_altrep <- vapply(materialized_altrep_classes, is.null, logical(1))
+
+    stopifnot(
+      "The subset array lengths do not match" = all(lengths(result_t[["array_subset"]]) == length(subset_indices)),
+      "The materialized array lengths do not match" = all.equal(lengths(result_t[["array_materialized"]]), array_lengths),
+      "Subset array was not materialized from altrep" = all(subset_is_altrep),
+      "Full array was not materialized from altrep" = all(materialized_is_altrep)
+    )
```
jonkeane, line 114: Nice, comprehensive checks here!

R/bm-array-altrep-materialization.R
```diff
@@ -0,0 +1,123 @@
...
+      )
+      table <- table[which(cols_without_nulls)]
+    }
+
+    # extract the arrays
+    arrays <- as.list(table)
+    array_lengths <- lengths(arrays)
+
+
+    # altrep checking
```
jonkeane, line 62:
```suggestion

    # altrep checking
```

Almost asinine whitespace nit

jonkeane (approved):
This looks fantastic. Would you mind running them locally + posting results of with | without I'm curious what kinds of results we get.

alistaire47:
> This looks fantastic. Would you mind running them locally + posting results of with | without I'm curious what kinds of results we get.

@jonkeane They're all very fast; this is really more of a microbenchmark. But generally materializing altreps is usually 2-4x slower than subsetting non-altrep, which by my intuition seems good given it's relative to subsetting, which should be extremely fast.

`type_integers`:

```tsv
   process     real start_mem_bytes end_mem_bytes max_mem_bytes source        exclude_nulls altrep subset_indices cpu_count
     <dbl>    <dbl>           <int>         <int>         <int> <chr>         <lgl>         <lgl>  <chr>              <int>
1 0.000393 0.000387       161890304     165478400     198524928 type_integers TRUE          FALSE  1:10                   1
2 0.000546 0.000539       161234944     164986880     197885952 type_integers TRUE          FALSE  1:10                  10
3 0.000962 0.000946       161480704     168214528     193970176 type_integers TRUE          TRUE   1:10                   1
4 0.00102  0.00100        167870464     179486720     201752576 type_integers TRUE          TRUE   1:10                  10
5 0.00266  0.00262        171540480     197246976     208142336 type_integers FALSE         FALSE  1:10                  10
6 0.00287  0.00282        168181760     193396736     204734464 type_integers FALSE         FALSE  1:10                   1
7 0.0130   0.0128         173309952     209682432     209682432 type_integers FALSE         TRUE   1:10                   1
8 0.0135   0.0141         171556864     213467136     213467136 type_integers FALSE         TRUE   1:10                  10
```

`fanniemae_2016Q4`:

```tsv
  process   real gc_level0 gc_level1 gc_level2 source           exclude_nulls altrep subset_indices cpu_count
    <dbl>  <dbl>     <int>     <int>     <int> <chr>            <lgl>         <lgl>  <chr>              <int>
1   0.406  0.400         0         0         2 fanniemae_2016Q4 TRUE          FALSE  1:10                   1
2   0.408  0.401         0         0         2 fanniemae_2016Q4 TRUE          FALSE  1:10                  10
3   2.13   2.10          0         0         6 fanniemae_2016Q4 TRUE          TRUE   1:10                  10
4   2.47   2.74          0         0         6 fanniemae_2016Q4 TRUE          TRUE   1:10                   1
5   3.85   5.74          0         0         3 fanniemae_2016Q4 FALSE         FALSE  1:10                  10
6   4.32   6.76          0         0         3 fanniemae_2016Q4 FALSE         FALSE  1:10                   1
7  10.6   14.3           0         0        14 fanniemae_2016Q4 FALSE         TRUE   1:10                  10
8  10.9   14.9           0         0        14 fanniemae_2016Q4 FALSE         TRUE   1:10                   1
```

jonkeane:
That's great, thanks!

PR #90, opened 2022-05-11
[ENG-3951] Add a json reading benchmark

### LINKED ISSUE #81: Add a json reading/writing benchmark
Analogous to: https://github.com/ursacomputing/arrowbench/blob/main/R/bm-write-csv.R and https://github.com/ursacomputing/arrowbench/blob/main/R/bm-read-csv.R

alistaire47:
Closes #81. Adds a benchmark to time `arrow::read_json_arrow()` and similar ndjson readers from other packages (jsonlite, ndjson, RcppSimdJson, and jsonify, though I disabled the jsonify for now due to instability). Added a writer to `ensure_format()` that uses `jsonlite::stream_out()`, which works, but is very slow (writing `fanniemae_2016Q4` takes hours), so we should replace it with an arrow writer if we write one.

Tests are wimpy because this is all quite slow.

Also renames the `output` param in `read_csv` (which this is heavily based upon) to `output_format`, because I realized it was getting overridden in the JSON by the captured output.


R/bm-read-json.R
```diff
@@ -0,0 +1,107 @@
...
+    }
+    con
+  }
+
+  reader_functions <- list(
+    arrow = arrow::read_json_arrow,
+    jsonlite = function(path, ...) {
+      con <- get_con(path)
+      on.exit(close(con))
+      jsonlite::stream_in(con = con)
```
boshek, line 88: If we add this, we won't get all that "Processing..." console output
```suggestion
      jsonlite::stream_in(con = con, verbose = FALSE)
```

R/ensure-format.R
```diff
@@ -141,8 +145,23 @@ get_write_function <- function(format, compression, chunk_size = NULL) {
...
+        con <- gzfile(path, open = "wb")
+      } else {
+        con <- file(path, open = "w")
+      }
+      on.exit(close(con))
+      x <- as.data.frame(x)
+      # TODO remove after data fixed; this makes stored dims not match
+      # remove null-type columns
+      x <- x[vapply(x, class, character(1L)) != 'vctrs_unspecified']
+      jsonlite::stream_out(as.data.frame(x), con = con, na = "null")
```
boshek, line 160: Probably worth doing here too
```suggestion
      jsonlite::stream_out(as.data.frame(x), con = con, na = "null", verbose = FALSE)
```

boshek, line 160: Probably worth doing here too
```suggestion
      jsonlite::stream_out(as.data.frame(x), con = con, na = "null", verbose = FALSE)
```

boshek (approved):
Looks good! Just noticed that the results were getting muddied a little by the jsonlite output so I think that is worth cleaning up. And wow does arrow really shine here!

R/ensure-format.R
alistaire47: Unless it speeds it up, I'd rather leave this one. Writing out fanniemae took hours, and this is the only way you can tell if it's hanging. The reading one is also slow, but because that gets captured you can't tell where it is anyway, so there's less point in keeping it. 

Long-term we need to replace jsonlite here with something faster.

alistaire47: Dupe comment? Confused how this happened

DESCRIPTION
```diff
@@ -12,29 +12,32 @@ License: MIT + file LICENSE
...
+    progress,
+    purrr,
+    remotes,
+    rlang,
+    R.utils,
+    sessioninfo,
+    utils,
+    waldo,
+    withr,
+    callr
```
jonkeane, line 28: 😬 oops, thanks for cleaning up this whitespace

R/bm-read-csv.R
```diff
@@ -13,10 +13,10 @@ read_csv <- Benchmark(
   setup = function(source = names(known_sources),
                    reader = "arrow",
                    compression = c("uncompressed", "gzip"),
-                   output = c("arrow_table", "data_frame")) {
+                   output_format = c("arrow_table", "data_frame")) {
     reader <- match.arg(reader, c("arrow", "data.table", "vroom", "readr"))
     compression <- match.arg(compression)
-    output <- match.arg(output)
+    output_format <- match.arg(output_format)
```
jonkeane, line 19: I'm totally fine with this change, though we should confirm that it won't bork ursacomputing/benchmarks calling of these.

https://github.com/ursacomputing/benchmarks/blob/f1114e0442de39239ca80d41b7e377598e4692dc/benchmarks/csv_benchmark.py#L44

R/bm-read-json.R
```diff
@@ -0,0 +1,107 @@
...
+    # NOTE: Removed jsonify from default options because it's unstable
+    jsonify = function(path, ..., as_data_frame) {
+      con <- get_con(path)
+      on.exit(close(con))
+      jsonify::from_ndjson(con, ...)
+    },
+    RcppSimdJson = function(path, ..., as_data_frame) {
+      lines <- readLines(path)
+      json <- RcppSimdJson::fparse(lines, single_null = NA, max_simplify_lvl = 3L, ...)
+      data.table::rbindlist(json)
```
jonkeane, line 100: I'm fine with keeping this here, though curious if there are other ways we might want to run this an rbind with either base or dplyr versions as well and if we should parameterize those (even if we don't test all of them all the time)

jonkeane (changes requested):
Looking good, a few comments + questions (though only one really substantive comment)

R/bm-read-json.R
alistaire47: The other packages handle rbinding to dataframe automatically (well, you can turn it off for {jsonlite} and do it manually if you want), so this is the only one that would get parameterized. {data.table} felt...appropriate? as also popular on the R-finance side and won't be arguable as not optimized enough. If we want we could parameterize for {RcppSimdJson} and {jsonlite} and I am curious what proportion of time is rbinding, but because {arrow} doesn't support not simplifying I wasn't sure to what extent we care. 

Also we have to be careful with `do.call(rbind, <ndjson list>)` because some of these do support implicit nulls (where the field just doesn't exist), which the base solution will screw up because of recycling. I only used explicit nulls in writing because it seems like a best practice anyway, but there's space here to screw it up if you pick the wrong combo. And arguably the lower level of robustness means the base solution is not quite equivalent.

R/bm-read-csv.R
alistaire47: Good call. It may keep working regardless because of partial matching? But we should change it anyway.

R/bm-read-json.R
jonkeane: > If we want we could parameterize for {RcppSimdJson} and {jsonlite} and I am curious what proportion of time is rbinding, but because {arrow} doesn't support not simplifying I wasn't sure to what extent we care.

Yeah, we probably don't care _that much_ ultimately — though being able to parameterize would tell us if {data.table} is fastest (I assume it is in this case, but it would be good to measure them all and check that). We can totally punt that until a later ticket too — it's a bit orthogonal to an extension anyway

alistaire47: I thought this would be pretty simple to add, but it turned out to be....a little not. The results are interesting, though; `dplyr::bind_rows()` is surprisingly slow. The base rbinding for {jsonlite} even includes an extra 

```r
lapply(out, function(x) lapply(x, function(y) if (is.null(y)) NA else y))
```
to replace `NULL` with `NA` (which it can't do on read, annoyingly, and `do.call(rbind.data.frame, ...)` won't work with NULLs), and it's _still_ faster than `dplyr::bind_rows()`. 

Also {jsonlite}-native simplification is faster than `data.table::rbindlist()`...maybe it can make stronger assumptions? Or maybe there's a lot of simplification overhead in {jsonlite} even if you try to turn it off? Not sure.

Using `type_integers`, uncompressed:

```tsv
   process   real gc_level0 gc_level1 gc_level2 reader       output_format rbinder    cpu_count
     <dbl>  <dbl>     <int>     <int>     <int> <chr>        <chr>         <chr>          <int>
 1   0.726  0.106         0         0         0 arrow        arrow_table   package           10
 2   0.775  0.108         0         0         0 arrow        data_frame    package           10
 3   4.62   4.55          3         3         7 RcppSimdJson data_frame    data.table        10
 4   6.90   6.81          4         2         7 ndjson       data_frame    package           10
 5   8.31   8.23         24         8         9 jsonlite     data_frame    package           10
 6  12.0   11.8          27         4         8 RcppSimdJson data_frame    base              10
 7  18.3   18.1         115        12        10 jsonlite     data_frame    data.table        10
 8  19.6   19.3          43         8        12 RcppSimdJson data_frame    dplyr             10
 9  31.1   30.6         143        17        13 jsonlite     data_frame    base              10
10  45.3   44.9         171        21        19 jsonlite     data_frame    dplyr             10
```

alistaire47: We could also try swapping out `readLines()` and `readr::read_lines()` or `vroom::vroom_lines()` for {RcppSimdJson} (and {jsonlite} if we like) if we want to try to give them the best chance possible, though it will make much less of a difference.

Also FWIW noticed in `?ndjson::stream_in()`:

> the package uses `data.table::rbindlist` for speed

jonkeane: Interesting, indeed!

R/bm-read-csv.R
alistaire47: PR: https://github.com/ursacomputing/benchmarks/pull/104

PR #91, opened 2022-05-11
ENG 3886: Save the script that is executed

### LINKED ISSUE #77: Save the script that is executed
Since we are using a subprocess to execute the benchmark, we should have a pretty clean way of generating a script that someone could run to get similar results locally. 

Let's surface that in the results for now + we can make an update to conbench to store + display that later

boshek:
Will close #77 once merged

This PR implements a `output_script` argument for `run_one` which defaults to TRUE and some tests. There is also some light cleanup around tests and documentation. Running this suite of benchmarks:

```r
run_benchmark(write_file, source = "nyctaxi_2010-01")
```

populates a "bm-scripts" directory with files that named similar to the results files except with an "rscript-" prefix. 
```bash
bm-scripts
└── write_file
    ├── rscript-lz4-1-feather-arrow_table-nyctaxi_2010-01.json
    ├── rscript-lz4-1-feather-data_frame-nyctaxi_2010-01.json
    ├── rscript-lz4-1-parquet-arrow_table-nyctaxi_2010-01.json
    ├── rscript-lz4-1-parquet-data_frame-nyctaxi_2010-01.json
    ├── rscript-lz4-10-feather-arrow_table-nyctaxi_2010-01.json
    ├── rscript-lz4-10-feather-data_frame-nyctaxi_2010-01.json
    ├── rscript-lz4-10-parquet-arrow_table-nyctaxi_2010-01.json
    ├── rscript-lz4-10-parquet-data_frame-nyctaxi_2010-01.json
    ├── rscript-snappy-1-parquet-arrow_table-nyctaxi_2010-01.json
    ├── rscript-snappy-1-parquet-data_frame-nyctaxi_2010-01.json
    ├── rscript-snappy-10-parquet-arrow_table-nyctaxi_2010-01.json
    ├── rscript-snappy-10-parquet-data_frame-nyctaxi_2010-01.json
    ├── rscript-uncompressed-1-feather-arrow_table-nyctaxi_2010-01.json
    ├── rscript-uncompressed-1-feather-data_frame-nyctaxi_2010-01.json
    ├── rscript-uncompressed-1-parquet-arrow_table-nyctaxi_2010-01.json
    ├── rscript-uncompressed-1-parquet-data_frame-nyctaxi_2010-01.json
    ├── rscript-uncompressed-10-feather-arrow_table-nyctaxi_2010-01.json
    ├── rscript-uncompressed-10-feather-data_frame-nyctaxi_2010-01.json
    ├── rscript-uncompressed-10-parquet-arrow_table-nyctaxi_2010-01.json
    └── rscript-uncompressed-10-parquet-data_frame-nyctaxi_2010-01.json
```

I am not entirely certain whether "bm-scripts" is best place or if that should be nested within "results" but I will wait for feedback on that.

### DIFF
--- a/tests/testthat/helper.R
+++ b/tests/testthat/helper.R
@@ -1,6 +1,3 @@
 library(jsonlite)
 
-wipe_results <- function() {
-  unlink(test_path("results/"), recursive = TRUE)
-  unlink(test_path("bm-scripts/"), recursive = TRUE)
-}
+wipe_results <- function() unlink(test_path("results/"), recursive = TRUE)
--- a/tests/testthat/helper.R
+++ b/tests/testthat/helper.R
@@ -1,0 +1,6 @@
+library(jsonlite)
+
+wipe_results <- function() {
+  unlink(test_path("results/"), recursive = TRUE)
+  unlink(test_path("bm-scripts/"), recursive = TRUE)
+}
--- a/tests/testthat/test-bm-row-group-size.R
+++ b/tests/testthat/test-bm-row-group-size.R
@@ -12,3 +12,5 @@ test_that("row_group_size benchmark runs", {
     "arrowbench_results"
   )
 })
+
+wipe_results()
--- a/.gitignore
+++ b/.gitignore
@@ -5,6 +5,7 @@
 tests/testthat/results
 .DS_Store
 results/
+bm-scripts/
 data/
 tests/testthat/data/
 .vscode/
--- a/.gitignore
+++ b/.gitignore
@@ -5,7 +5,6 @@
 tests/testthat/results
 .DS_Store
 results/
-bm-scripts/
 data/
 tests/testthat/data/
 .vscode/
--- a/R/run.R
+++ b/R/run.R
@@ -14,6 +14,8 @@
 #' `profvis::profvis(prof_input = file)`. Default is `FALSE`
 #' @param read_only this will only attempt to read benchmark files and will not
 #' run any that it cannot find.
+#' @param output_script logical: output the rscript used to run the benchmark as a
+#' json file
 #'
 #' @return A `arrowbench_results` object, containing a list of length `nrow(params)`,
 #' each of those a `list` containing "params" and either "result" or "error".
@@ -27,7 +29,8 @@ run_benchmark <- function(bm,
                           n_iter = 1,
                           dry_run = FALSE,
                           profiling = FALSE,
-                          read_only = FALSE) {
+                          read_only = FALSE,
+                          output_script = TRUE) {
   start <- Sys.time()
   stopifnot(is.data.frame(params))
   message("Running ", nrow(params), " benchmarks with ", n_iter, " iterations:")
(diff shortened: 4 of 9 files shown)


tests/testthat/helper.R
```diff
@@ -0,0 +1,6 @@
+library(jsonlite)
+
+wipe_results <- function() {
+  unlink(test_path("results/"), recursive = TRUE)
+  unlink(test_path("bm-scripts/"), recursive = TRUE)
```
alistaire47, line 5: This implies the scripts are part of the results, and should probably be in the same directory, no? Not sure what's ideal here, but maybe an `outputs` dir with `results` and `scripts` subdirs?

alistaire47:
I'm confused why we're storing scripts in JSON. Do we want them to be machine-readable? Are plain R files less so? I get this avoids needing subdirectories for each benchmark to contain setup/run/teardown/whatever scripts, but that's much easier to navigate as a human, and not that hard to navigate as a computer.

I guess that reaction is because when I see a `scripts` directory that just contains a lot of JSON I get confused

jonkeane:
Nice work! I'll have more substantive comments later (and maybe we can pair on this when we're together too...). But this is going in the right direction. 

I totally agree that it's weird to have these scripts in json, but we will (eventually) be sending them up to conbench (once conbench can accept those...) so json makes sense. I was originally thinking that we would actually put that json inside of the results json (so that, for example, you could get the lines of the script from the results object) as opposed to a separate json file — do you think that's doable?

boshek:
> do you think that's doable

Very. That's a much cleaner idea. I think it also removes the need to have an argument for whether to include the script or not. Probably just easier to assume it will be there.

tests/testthat/helper.R
boshek: removed now

tests/testthat/helper.R
```diff
@@ -0,0 +1,3 @@
+library(jsonlite)
+
+wipe_results <- function() unlink(test_path("results/"), recursive = TRUE)
```
alistaire47, line 3: Do we want to be more targeted with how we use this? We could remove specific results when we need, or everything on teardown and/or setup, if we like. I worry we're going to have to go figure out where this is getting called and comment it out if we want to see the results for debugging.

Also jsonlite is not necessary anymore

tests/testthat/test-bm-row-group-size.R
```diff
@@ -12,3 +12,5 @@ test_that("row_group_size benchmark runs", {
     "arrowbench_results"
   )
 })
+
+wipe_results()
```
alistaire47, line 16: I don't think this should be in some benchmarks but not all. If it's going to affect all, it should be at the top level where one call handles everything.

tests/testthat/helper.R
boshek: Ok @jonkeane is doing this here: https://github.com/ursacomputing/arrowbench/pull/92

PR #92, opened 2022-05-12
Only install duckdb when you really need it

jonkeane:


DESCRIPTION
```diff
@@ -13,6 +13,7 @@ Encoding: UTF-8
 Depends: R (>= 3.5.0)
 Imports:
   bench,
+  dbplyr,
```
jonkeane, line 16: This "shouldn't" be necessary, but in order to get it installed on CI and running it's needed. We can refactor this out, but I would prefer to defer it for now

R/bm-tpc-h.R
```diff
@@ -114,36 +114,6 @@ tpc_h <- Benchmark("tpc_h",
...
-      for (col in cols) {
-        answer[, col] <- as.Date(answer[, col])
-      }
-
-      all_equal_out <- waldo::compare(result, answer, ignore_attr = TRUE, tolerance = 0.01)
-      if (length(all_equal_out) != 0) {
-        warning(paste0("\n", all_equal_out, "\n"))
-        stop("The answer does not match")
-      }
-    }
```
jonkeane, line 145: This was always a bit overly cautious since these also get answers we have here.

alistaire47 (approved):

PR #98, opened 2022-05-24
Make result JSON self-contained

### LINKED ISSUE #97: Make result JSON self-contained
In the effort towards moving {benchmarks} from using `run_one()` to `run_benchmark()` (and presumably eventually moving to a more abstracted orchestrator), the result JSON produced by {arrowbench} should contain all metadata necessary to post it to conbench—everything called `context`, `info`, `tags`, etc. on the {benchmarks} side. 

This task is to rearrange as required such that a directory of JSONs produced by a run can be hoovered up and sent to conbench with no additional context or metadata.

alistaire47:
Closes #97. An effort to add metadata to the JSON generated by {arrowbench} such that it contains everything required to pass to `conbench.record()` (like [happens in {ursacomputing/benchmarks} right now](https://github.com/ursacomputing/benchmarks/blob/main/benchmarks/_benchmark.py#L103-L112)) such that a directory of results can be handled without maintaining metadata elsewhere, a prerequisite for switching from using `run_one()` to `run_benchmark()` in {ursacomputing/benchmarks}.

Notes:

- Adds `auto_unbox = TRUE, null = 'null'` to `toJSON()` options because `{}` is not `NULL` and we don't want lists of length 1 everywhere. This doesn't seem to break anything.
- Hardcodes `drop_caches = FALSE` for now, as that's the default in {benchmarks} and we're not overriding it anywhere AFAIK. We should probably clear them from R and parameterize this, but that felt beyond the scope of this story.
- Adds the metadata as attributes to the dataframe form of results because they were getting too wide anyway. Could probably chuck some of the other nonvarying fields in there too, but didn't want to break anything.
- This ignores what are called `extra_tags` in {benchmarks} because I don't see them used anywhere. 
- I didn't mess with `output` (`stdout` for the call), though I think this gets handled differently depending on where it gets called. We'll likely need to revisit how we want it to work.
- This schema should be reasonably complete, but it could use some standardization—what `benchmarks.Conbench.record()`, `conbench.record()`, and the conbench POST API want are all slightly different. There is also duplication in here; notably params for the case are both in `params` and `tags`. A pydantic schema we use everywhere (and recreate in R6?) would be nice.

What the JSON now looks like (lightly abridged):

```json
{
  "name": "array_altrep_materialization",
  "tags": {
    "exclude_nulls": false,
    "altrep": false,
    "subset_indices": [
      [1, 2, 3, 4, 5, 6, 7, 8, 9, 10]
    ],
    "dataset": "type_integers",
    "language": "R"
  },
  "info": {
    "arrow_version": "9.0.0-SNAPSHOT",
    "arrow_compiler_id": "AppleClang",
    "arrow_compiler_version": "13.1.6.13160021",
    "benchmark_language_version": "R version 4.1.3 (2022-03-10)",
    "arrow_version_r": "8.0.0.9000"
  },
  "context": {
    "arrow_compiler_flags": " -Qunused-arguments -fcolor-diagnostics -O3 -DNDEBUG",
    "benchmark_language": "R"
  },
  "github": {
    "repository": "https://github.com/apache/arrow",
    "commit": "d9d78946607f36e25e9d812a5cc956bd00ab2bc9"
  },
  "options": {
    "iterations": 1,
    "drop_caches": false,
    "cpu_count": null
  },
  "result": [
    {
      "process": 0.00262000000000001,
      "real": 0.0025826720520854,
      "start_mem_bytes": 178601984,
      "end_mem_bytes": 192430080,
      "max_mem_bytes": 215236608,
      "gc_level0": 0,
      "gc_level1": 0,
      "gc_level2": 0
    }
  ],
  "params": {
    "source": "type_integers",
    "exclude_nulls": false,
    "altrep": false,
    "subset_indices": [
      [1, 2, 3, 4, 5, 6, 7, 8, 9, 10]
    ],
    "cpu_count": 10,
    "lib_path": "latest",
    "packages": [
      {
        "package": "arrow",
        "version": "8.0.0.9000",
        "date": "2022-05-12",
        "source": "local"
      },
      {
        "package": "arrowbench",
        "version": "0.1.0",
        "date": "2022-05-24",
        "source": "local"
      },
      ... < this gets very long >
    ]
  },
  "output": "### RESULTS HAVE BEEN PARSED ###",
  "rscript": [
    "",
    "library(arrowbench)",
    ...
    "cat(\"\n##### RESULTS END\n\")"
  ]
}
```


alistaire47:
Maybe we should separate the case params from the global params (`cpu_count`, `lib_path`, and `packages`). Doesn't have to happen now and I haven't yet seen what downstream effects would be, but I feel like sooner or later conflating args and state is going to bite us, and now we have lots of metadata describing state anyway.

R/run.R
```diff
@@ -186,14 +186,29 @@ run_bm <- function(bm, ..., n_iter = 1, profiling = FALSE, global_params = list(
...
-  params$packages <- package_info()[, c("package", "loadedversion", "date", "source")]
-  names(params$packages)[2] <- "version"
-
-  structure(list(
-    result = do.call(rbind, results),
-    params = params
-  ), class = "arrowbench_result")
+  all_params <- modifyList(params, global_params)
+  all_params$packages <- package_info()[, c("package", "loadedversion", "date", "source")]
+  names(all_params$packages)[2] <- "version"
```
boshek, line 191: we can assume that "loadedversion" will always be at this position right?

boshek (approved):
This looks really good to me and I don't really have any changes to the code. Really great work. Three thoughts:
- given changes like this I wonder if we should consider versioning arrowbench a little more rigorously. i suppose this will hopefully be a one time change but when we coalesce on a common schema it might be worth tagging a release
- your thought on separating state from args is a good one but...
- does this PR need a decision point before merging? I know it'll work as is but should all the schema pieces be sorted out before then?

R/run.R
```diff
@@ -186,14 +186,29 @@ run_bm <- function(bm, ..., n_iter = 1, profiling = FALSE, global_params = list(
...
+  row.names(all_params$packages) <- NULL
+  # remove `packages_info` class whose print method won't work anymore
+  all_params$packages <- as.data.frame(all_params$packages)
```
jonkeane, line 194: 🙏 this was bugging me elsewhere recently too

R/run.R
```diff
@@ -241,6 +256,56 @@ global_setup <- function(lib_path = NULL, cpu_count = NULL, mem_alloc = NULL, te
...
+  info <- list(
+    arrow_version = arrow_info$build_info$cpp_version,
+    arrow_compiler_id = arrow_info$build_info$cpp_compiler,
+    arrow_compiler_version = arrow_info$build_info$cpp_compiler_version,
+    benchmark_language_version = version[['version.string']],
+    arrow_version_r = as.character(arrow_info$version)
+  )
+
+  context <- list(
+    arrow_compiler_flags = arrow_info$build_info$cpp_compiler_flags,
```
jonkeane, line 284: Until I saw this here, I hadn't totally put together how much this could actually differ from what we see in conbench (depending on how exactly we install the arrow R package, if we install pointing to libarrow that was build during orchestration it'll be the same, but if we install with a static lib, these flags could actually be quite different than the ones we see from the conbench|benchmarks utilities we are using now). Very nice!

R/run.R
```diff
@@ -241,6 +256,56 @@ global_setup <- function(lib_path = NULL, cpu_count = NULL, mem_alloc = NULL, te
...
+    benchmark_language = "R"
```
jonkeane, line 285: We might need this for now, but I wonder if this + language above are signs that we should generalize these in the json? We don't have to do that in this PR of course, but a though for when we're designing canonical conbench json

R/result.R
```diff
@@ -23,5 +23,13 @@ as.data.frame.arrowbench_result <- function(x, row.names = NULL, optional = FALS
   }
   out <- cbind(iteration = seq_len(nrow(x$result)), x$result, x$params)
   out$output <- x$output
+
+  metadata_elements <- c("name", "tags", "info", "context", "github", "options")
+  for (element in metadata_elements) {
+    if (element %in% names(x)) {
+      attr(out, element) <- x[[element]]
+    }
+  }
```
jonkeane, line 32: Could you add a comment describing what this is doing?

jonkeane (changes requested):
Looking great, thanks! 

Mostly questions + comments inspired by reading this code. The only substantive thing I would consider adding is: how hard would it be to make a reader function in {arrowbench} that constructs a results object from one of these JSONs (without being run through `run_benchmark()`)? That would let us experiment with this a bit more easily in {arrowbench} (and be the start of the class for results objects we've talked about)

R/run.R
alistaire47: Should be ok because of the previous line. I don't love the hardcoded index, but this is only part of the diff because I changed the variable name. Open to a cleaner alternative

R/run.R
alistaire47: Yeah I was getting errors when trying to print and wasn't really sure if it was my fault, so I just fixed it regardless

R/run.R
alistaire47: Yeah, I think a lot of this is ripe for rearranging, but was just trying to recreate what the Python already generates for the moment. Big picture we're currently slightly constrained by the top-level fields in conbench, behind which I didn't see a lot of reasoning. Didn't see != doesn't exist, but from looking at the conbench code a lot I do think it's worth a conversation if we come up with a well-thought-out alternative, maybe with `language` at/near the top level

alistaire47:
> The only substantive thing I would consider adding is: how hard would it be to make a reader function in {arrowbench} that constructs a results object from one of these JSONs (without being run through `run_benchmark()`)? That would let us experiment with this a bit more easily in {arrowbench} (and be the start of the class for results objects we've talked about)

For a reasonably comprehensive implementation (I'm thinking R6, maybe light validation, adjusting the `as.data.frame` methods, serialization to/from JSON, and adjusting all the code to use it everywhere) nontrivial, but doable. Maybe 2-3 days? A little more if something's more complicated than I realize.

jonkeane:
> For a reasonably comprehensive implementation (I'm thinking R6, maybe light validation, adjusting the as.data.frame methods, serialization to/from JSON, and adjusting all the code to use it everywhere) nontrivial, but doable. Maybe 2-3 days? A little more if something's more complicated than I realize.

Cool, let's do that as a follow on, though do that relatively soon (excepting time off + the weekend etc. of course!)

PR #103, opened 2022-06-02
Add fuller benchmark results class

### LINKED ISSUE #99: Add fuller benchmark results class
Per https://github.com/ursacomputing/arrowbench/pull/98#issuecomment-1137898948, create an object for a set of benchmark results (at the `run_one()` level, i.e. one set of scalar params (a case), perhaps with multiple iterations) that can be read/written from/to JSON. Possibilities in-scope:

- R6 
- light validation
- adjusting the as.data.frame methods
- serialization to/from JSON
- adjusting all the code to use it everywhere

Exhaustive validation is out of scope for now, given part of the point is to allow us to play with schemas more.

alistaire47:
Closes #99. Implements some more robust R6 classes in `result.R` to replace the old S3 `arrowbench_result` and `arrowbench_results`. The new classes can easily be serialized/deserialized to/from lists/strings of JSON/JSON files, and have active bindings for the various attributes so there's a clear place in the code to add validation for each that will get called regardless of how the instance is created (e.g. complete with `$new()`, deserialized from JSON, or overriding various bits later). This adds some boilerplate now, but means we don't need to rearrange anything to add validation later.

The flow for the `as.data.frame()` methods and `get_params_summary()` has been rearranged to call R6 methods, so you can call them directly, but existing code will continue to work exactly the same.

This PR notably does _not_ actually implement the validation it enables, given we're likely to evolve the schemas this defines soon. If we eventually want to make a nested `ResultMetadata` or `Packages` object that lets us easily validate further down, the tools are in place though (e.g. recursive serialization should mostly just work).


R/result.R
```diff
@@ -1,36 +1,343 @@
+# Merge two named vectors, overwriting identically-named elements of x1 with
+# that of x2.
+coalesce_vectors <- function(x1, x2) {
+  out <- x1
+  to_overwrite <- names(x2) %in% names(x1)
+  out[names(x2)[to_overwrite]] <- x2[to_overwrite]
+  out <- c(out, x2[!to_overwrite])
+  out
+}
```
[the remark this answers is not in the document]
alistaire47: It's basically the same, except the current version works with atomic vectors or general vectors (lists), keeping the type. But we only really care about lists here, so actually I think I can delete all this and just use `modifyList()` itself

R/result.R
```diff
@@ -1,36 +1,343 @@
...
+        private$.output <- output
+      }
+      private$.output
+    },
+    rscript = function(rscript) {
+      if (!missing(rscript)) {
+        private$.rscript <- rscript
+      }
+      private$.rscript
+    },
```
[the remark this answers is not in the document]
alistaire47: K, added a `private$get_or_set()` method to make this more compact for now. Left it as a single line for each rather than trying to generate the list; this way it's easy to see what's there as well as change them if/when we want to add validation.

alistaire47:
> The only other thing that springs to mind right now is: should we also have a test for the roundtrip write results to disk, and then read them back in as if this were a new session | the results object wasn't coming directly from the benchmark running?

The existing usage does test a bunch of paths already because of how it works across processes (so when you screw something up here _everything_ breaks 😅), but yeah, it's a good idea to have some isolated unit tests that just instantiate manually and test directly. I'll add `test-result.R`

alistaire47:
Added a bunch of unit tests; could add more if it's useful. Are we tracking coverage?

alistaire47:
CI failures look semi-random and maybe related to #92. Don't really understand what's going on yet, though.

alistaire47:
Ah crap it's a `stringsAsFactors` issue on R 3.x

R/result.R
```diff
@@ -1,36 +1,343 @@
...
+  out <- c(out, x2[!to_overwrite])
+  out
+}
+
+
+# Version of R6 with heritable static/class methods and attributes
+#
+# Elements in `static` can be called without instantiation, e.g. `Class$method()`.
+# Functions are evaluated in the environment of `Class`, so you can refer to `self`
+# (which is the class—not the instance—here) to create class methods.
```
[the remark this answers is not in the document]
boshek: I wonder if some version of this might be useful in the README (or if that is too long in people's mind a vignette). This is a really good explanation and I think it could be useful to explain how/why one might approach using this.

R/result.R
```diff
@@ -1,36 +1,285 @@
...
+    name = function(name) private$get_or_set(variable = ".name", value = name),
+    result = function(result) private$get_or_set(variable = ".result", value = result),
+    params = function(params) private$get_or_set(variable = ".params", value = params),
+    tags = function(tags) private$get_or_set(variable = ".tags", value = tags),
+    info = function(info) private$get_or_set(variable = ".info", value = info),
+    context = function(context) private$get_or_set(variable = ".context", value = context),
+    github = function(github) private$get_or_set(variable = ".github", value = github),
+    options = function(options) private$get_or_set(variable = ".options", value = options),
+    output = function(output) private$get_or_set(variable = ".output", value = output),
+    rscript = function(rscript) private$get_or_set(variable = ".rscript", value = rscript),
```
boshek, line 171: Sort of unrelated to this but for validation will our best approach be to define validation functions a la:
```r
    name = function(name) is_ab_name(private$get_or_set(variable = ".name", value = name)),
    result = function(result) is_ab_result(private$get_or_set(variable = ".result", value = result))
    ...
```
Question more for my understanding than anything.

R/result.R
```diff
@@ -1,36 +1,285 @@
...
+
+  public = list(
+    initialize = function(results) {
+      self$results <- results
+    },
+    data.frame = function(row.names = NULL, optional = FALSE, ...) {
+      x <- self$results
+      valid <- map_lgl(x, ~inherits(.x, "BenchmarkResult"))  # failures will be BenchmarkFailure
+
+      dplyr::bind_rows(lapply(x[valid], function(res) res$data.frame(...)))
```
boshek, line 248: Is there a reason not to make this a tibble? We use data.frames and tibbles interchangeably. Should we standardize?

tests/testthat/test-result.R
```diff
@@ -0,0 +1,74 @@
...
+  expect_identical(SumOtherClass$sum, sum)
+
+  expect_equal(SumOtherClass$sum(SumOtherClass$x), 5050L)
+})
+
+
+test_that("inherited serialization/deserialization methods work", {
+  res <- BenchmarkResult$new(
+    name = "fake",
+    result = data.frame(time = 0, status = "superfast"),
```
boshek, line 27: 🏎️

boshek (approved):
This is really impressed work. A couple small comments and two things to consider:

1) I think documenting this a little more fully would be good
2) As you suggested, I do wonder if a parent class for results and errors might be useful. Just surfacing that comment. I can't think of a specific need but just wonder if generally might be good.

R/result.R
alistaire47: Nope, we can't do that because the validation really needs to happen in the middle of that function, before it assigns with `private$.name <- name`. So probably we'll need to revert it to 

``` r
name = function(name) {
  if (!missing(name)) {
    # validation starts here, e.g.
    stopifnot(is_ab_name(name))
    # validation ends here
    private$.name <- name
  }
  private$.name
},
```

I thought about making `get_or_set()` take an expression or function for validation, but I don't want it to live in yet another location, so if we go that route, it would need to be defined in the call, e.g.

``` r
name = function(name) private$get_or_set(
  variable = ".name", 
  value = name, 
  validation = {stopifnot(is_ab_name(name))}
),
```
or something like that. But currently I prefer the more verbose version without `get_or_set()` because it's more transparent.

R/result.R
alistaire47: It is because `BenchmarkResult$new(...)$data.frame(...)` returns a tibble. I just named the method `$data.frame()` so it's parallel with the `BenchmarkResult$new(...)$list` and `$json` active bindings. But really it's a method, not an active binding, so maybe it should be called `$to_dataframe()` or something? Also `data.frame` vs. `dataframe` vs. `data_frame` vs. `data frame` vs. `tibble` is a mess...I don't know what the most intuitive is from a user perspective.

tl;dr: kind of want to change this name; no idea what to change it to; suggestions welcome

boshek: Ah I see. Maybe `to_tibble` or `bm_tibble` might be good too.

alistaire47: Yeah, I'm leaning towards `$to_dataframe()` or `$to_tibble()`

alistaire47:
@austin3dickey It won't let me make you a reviewer on this, but please have a look through the serialization/deserialization bits if you've got a minute

R/result.R
```diff
@@ -1,36 +1,285 @@
+# Version of R6 with heritable static/class methods and attributes
+#
+# Elements in `static` can be called without instantiation, e.g. `Class$method()`.
+# Functions are evaluated in the environment of `Class`, so you can refer to `self`
+# (which is the class—not the instance—here) to create class methods.
+#' @include util.R
+R6.1Class <- function(..., static = NULL) {
```
austin3dickey, line 7: This is really nitpicky; feel free to disregard. But I've been trying to grok this for a while as "the R6 method of 1Class classes," in the style of e.g. `summary.Date()` or whatever. Part of that is definitely R's fault in overloading `m.c` to mean either "the m method of c classes" or "the m_c variable," exacerbated by the fact that I haven't read R in over a year, but maybe it would be a bit clearer to call this something else? I still like the 6.1 reference, so maybe `R6Point1Class`?

alistaire47: Oh yes I like that

jonkeane (approved):
I'm good with these changes, thanks!

R/result.R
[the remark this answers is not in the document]
alistaire47: Added a pile more docs, including a version of the above

R/result.R
```diff
@@ -1,36 +1,364 @@
+#' Version of R6 with heritable static/class methods and attributes
```
austin3dickey, line 1: This is really cool, Ed. Honestly I think this and the `Serializable` deserve their own package, or at least issue in https://github.com/r-lib/R6 .

austin3dickey: Maybe `Serializable` doesn't necessarily belong in R6, but heritable classmethods are such a common tool in Python that I'm surprised they don't already exist in R6.

R/result.R
```diff
@@ -1,36 +1,364 @@
...
+  Class <- R6::R6Class(...)
+  Class$parent_env <- parent.frame()
+
+  full_static <- modifyList(
+    Class$get_inherit()$private_fields$static %||% list(),
+    static %||% list()
+  )
+  full_static <- lapply(full_static, function(x) {
+    if (!is.function(x)) return(x)
+    environment(x) <- Class
```
austin3dickey, line 58: If you wanted to get really fancy, you could rename `self` to `cls` or something here. Though I've never really liked the "reserved" nature of the keyword `self` in R6 so maybe that's not a great idea to add to that cognitive overhead

R/result.R
```diff
@@ -1,36 +1,364 @@
...
+#
+# Specifically:
+# - `$json`: Active binding that returns a JSON string
+# - `$list`: Active binding that returns a list
+# - `$write_json()`: Method that writes a JSON representation to a file
+# - `$from_json()`: Class method that creates an instance from a JSON string
+# - `$from_list()`: Class method that creates an instance from a list
+# - `$read_json()`: Class method that creates an instance from a JSON file
+#
+# Note that private attributes with a `.` prefix will be serialized to JSON
```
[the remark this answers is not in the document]
austin3dickey: I like `private$to_serialize <- list(...)` (this is how I used to handle this problem I think). The code to serialize and deserialize will be a little cleaner than parsing the names, and you can avoid the confusing dot notation.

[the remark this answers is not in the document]
austin3dickey: I was just gonna say that! Nice!

austin3dickey (approved):
nice job!!

R/result.R
alistaire47: Yeah, I considered that; it's weird to have a `self` which is not quite the same. But it is still called `self` by R6 itself...it was clearly just not quite designed to do what I'm doing. For now I think I have a mild preference for leaving it as-is since it will make the names match when digging into classes when debugging.

PR #104, opened 2022-06-07
Store a marker for benchmark versions

### LINKED ISSUE #101: Store a marker for benchmark versions
Per https://github.com/conbench/conbench/issues/314, add versioning to benchmark and their results

alistaire47:
Closes #101. Adds semantic versioning to benchmarks that can be incremented when benchmarks are altered. Alters JSON schema for results from

```json
{
  "name": "array_altrep_materialization",
  ...
}
```

to 

```json
{
  "benchmark": {
    "name": "array_altrep_materialization",
    "version": "1.0.0"
  },
  ...
}
```

as name and version together identify a benchmark. All versions are set to `1.0.0` for now.

These changes have not been implemented in conbench or ursacomputing/benchmarks yet, but since {benchmarks} is ignoring everything in the JSON except `blob.result.real`, this shouldn't break anything. However, {benchmarks} and {conbench} should be brought in line with this behavior; if we make this change, https://github.com/conbench/conbench/issues/314 will require simultaneous PRs to both {conbench} and {benchmarks} so as not to break anything.

### DIFF
--- a/R/benchmark.R
+++ b/R/benchmark.R
@@ -63,6 +63,8 @@
 #' the `teardown()` function.
 #'
 #' @param name string identifier for the benchmark, included in results
+#' @param version string or numeric version object (as returned by
+#' `numeric_version()`) of semantic version for the benchmark
 #' @param setup function having as its arguments the benchmark parameters. See
 #' the `Parametrizing benchmarks` section. This function is called once
 #' to initialize the benchmark context for a given set of parameters.
@@ -87,6 +89,7 @@
 #' @return A `Benchmark` object containing these functions
 #' @export
 Benchmark <- function(name,
+                      version,
                       setup = function(...) BenchEnvironment(...),
                       before_each = TRUE,
                       run = TRUE,
@@ -98,6 +101,7 @@ Benchmark <- function(name,
   structure(
     list(
       name = name,
+      version = numeric_version(version, strict = TRUE),
       setup = setup,
       before_each = substitute(before_each),
       run = substitute(run),
--- a/R/bm-array-to-vector.R
+++ b/R/bm-array-to-vector.R
@@ -10,7 +10,10 @@
 #'
 #' @importFrom purrr map flatten
 #' @export
-array_to_vector <- Benchmark("array_to_vector",
+array_to_vector <- Benchmark(
+  "array_to_vector",
+  version = "1.0.0",
+
   setup = function(
       # the only datasets that have any no-null numerics are
       source = c("type_integers", "type_floats"),
--- a/tests/testthat/test-result.R
+++ b/tests/testthat/test-result.R
@@ -23,7 +23,7 @@ test_that("R6.1 classes inherit properly", {
 
 test_that("inherited serialization/deserialization methods work", {
   res <- BenchmarkResult$new(
-    name = "fake",
+    benchmark = list(name = "fake", version = "1.0.0"),
     result = data.frame(time = 0, status = "superfast", stringsAsFactors = FALSE),
     params = list(speed = "lightning"),
     tags = c(is_real = FALSE)
@@ -31,7 +31,7 @@ test_that("inherited serialization/deserialization methods work", {
 
   # sanity
   expect_s3_class(res, "BenchmarkResult")
-  expect_equal(res$name, "fake")
+  expect_equal(res$benchmark, list(name = "fake", version = "1.0.0"))
 
   # roundtrips
   expect_equal(res$json, BenchmarkResult$from_json(res$json)$json)
@@ -45,7 +45,7 @@ test_that("inherited serialization/deserialization methods work", {
 
 test_that("S3 methods work", {
   res <- BenchmarkResult$new(
-    name = "fake",
+    benchmark = list(name = "fake", version = "1.0.0"),
     result = data.frame(time = 0, status = "superfast", stringsAsFactors = FALSE),
     params = list(speed = "lightning"),
     tags = c(is_real = FALSE)
@@ -60,7 +60,7 @@ test_that("S3 methods work", {
     structure(
       list(iteration = 1L, time = 0, status = "superfast", speed = "lightning"),
       row.names = c(NA, -1L), class = c("tbl_df", "tbl", "data.frame"),
-      name = "fake", tags = c(is_real = FALSE)
+      benchmark = list(name = "fake", version = "1.0.0"), tags = c(is_real = FALSE)
     )
   )
 
--- a/R/bm-array-altrep-materialization.R
+++ b/R/bm-array-altrep-materialization.R
@@ -12,6 +12,7 @@
 #' @export
 array_altrep_materialization <- Benchmark(
   "array_altrep_materialization",
+  version = "1.0.0",
 
   setup = function(source = names(known_sources),
                    exclude_nulls = FALSE,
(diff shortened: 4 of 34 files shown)


tests/testthat/test-result.R
```diff
@@ -45,7 +45,7 @@ test_that("inherited serialization/deserialization methods work", {
 
 test_that("S3 methods work", {
```
austin3dickey, line 46: wanna add a test if the `benchmark` object is malformed? there are a lot of rules there

R/bm-array-to-vector.R
```diff
@@ -10,7 +10,10 @@
 #'
 #' @importFrom purrr map flatten
 #' @export
-array_to_vector <- Benchmark("array_to_vector",
+array_to_vector <- Benchmark(
+  "array_to_vector",
+  version = "1.0.0",
```
jonkeane, line 15: This might be too complicated to implement nicely, but one thing that this will do is keep the version for each case of the benchmark. And in practice I bet that most situations where we bump versions will actually be for a single case of a benchmark rather than all of them together. Any thoughts about it that's something we should account for?

I'm not really concerned about the additional versions (the numbers are free, after all!), but version bumps will + should disrupt history, and so we don't want to do that unless we actually need to, yeah?

jonkeane (changes requested):
The code here looks good, though like I mention in comments: I wonder if there's anyway we could actually version at the case level?

One more general comment: any reason for using semantic versions? I'm generally pro semantic versions, but for these I'm not sure we will ever need to distinguish between X Y or Z in `X.Y.Z` to treat those differently.

R/benchmark.R
```diff
@@ -87,6 +89,7 @@
 #' @return A `Benchmark` object containing these functions
 #' @export
 Benchmark <- function(name,
+                      version,
```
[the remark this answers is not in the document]
jonkeane: Oh, and, do we want to have this as an argument here or something that's set inside the body of the benchmarking code? We don't expect that we will support both versions such that you can select one or the other, right?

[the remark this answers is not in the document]
alistaire47: I didn't give it a default because I wanted the version to necessarily live in each benchmark so it's more visible when it should be updated. It's an extra line, but it feels worth it? Otherwise we (at least I) will likely forget to change these when we alter benchmarks

alistaire47:
> One more general comment: any reason for using semantic versions? I'm generally pro semantic versions, but for these I'm not sure we will ever need to distinguish between X Y or Z in `X.Y.Z` to treat those differently.

I was thinking about this and initially was leaning towards just an integer, but really we can/do make smaller non-breaking changes to benchmarks. Maybe those aren't worth versioning if it's just to cause a history discontinuity? If so, we need to document thoroughly when versions _should_ be incremented and what the implications are.

tests/testthat/test-result.R
alistaire47: yeah, a good idea. it's overkill really; `numeric_version()` already does a lot of checking for us

jonkeane:
> I was thinking about this and initially was leaning towards just an integer, but really we can/do make smaller non-breaking changes to benchmarks. Maybe those aren't worth versioning if it's just to cause a history discontinuity? If so, we need to document thoroughly when versions should be incremented and what the implications are.

Yeah, it's possible that there are other reasons we might want to version that won't reset history, but I honestly can't think of any off the top of my head. Agreed that either way we should document that version bumps (or which version number bumps) cause history to be reset.

alistaire47:
> Yeah, it's possible that there are other reasons we might want to version that won't reset history, but I honestly can't think of any off the top of my head.

I'm thinking about how the altrep materialization one has a `query` param where most queries (aside from `everything`) are specific to the dataset, and not all datasets have queries implemented. New queries would create new cases; I'm not sure if that _should_ break history? I'd think of it as an additive minor version bump. Similar for when Sam added compression to CSV writing.

For patch versions I think I've done some minor refactoring which might count?

In all cases I don't know if we care—do these changes merit versioning? Or do we only care about changes that will impact times for a particular case? If the latter that's fine, we just need to make sure we don't increment when we shouldn't.

jonkeane:
> In all cases I don't know if we care—do these changes merit versioning? Or do we only care about changes that will impact times for a particular case? If the latter that's fine, we just need to make sure we don't increment when we shouldn't.

Yeah, I don't think we care for those. I mean we have a version of sorts in the git history — though that's not connected or recorded in conbench, I think that's not really important. I could totally be convinced we do want to save this information, but so far I haven't really seen a reason.

> I'm thinking about how the altrep materialization one has a query param where most queries (aside from everything) are specific to the dataset, and not all datasets have queries implemented.

This is a bit tangential to a slightly different problem: (and I swear I had an issue for this somewhere, but can't find it now!) for some benchmarks we use subsets of sources, but we don't really record that as part of the source (it's frequently an independent option in the cases).

R/bm-array-to-vector.R
jonkeane: Hmmm, yeah that might work. I was also thinking separately, what if by default we ran version "highest" and then we defined in the actual benchmark code which case(s) have which numbers? 

So by default everything starts out as 1. Then if we change something for some specific case, we do something like we do with the validity bit and version 2 for that one case is selected when you run "highest" (we could decide separately if we want to keep around the code in if|else for running older versions too then)

alistaire47:
> This is a bit tangential to a slightly different problem: (and I swear I had an issue for this somewhere, but can't find it now!) for some benchmarks we use subsets of sources, but we don't really record that as part of the source (it's frequently an independent option in the cases). But I bet we could do something more reliable + faithful to what's going on that isn't custom to each benchmark

This sounds potentially complicated; the reasons I've seen for subsetting get complicated fast, e.g. the altrep one subsets to types that can be represented by altrep. The other messy one I've seen is all the mess with the Fannie Mae null columns, but we should just fix those (for which we do have a story)

jonkeane:
> This sounds potentially complicated; the reasons I've seen for subsetting get complicated fast, e.g. the altrep one subsets to types that can be represented by altrep. The other messy one I've seen is all the mess with the Fannie Mae null columns, but we should just fix those (for which we do have a story)

Totally, and maybe ultimately any generic subsetting we do should be built into the data sources tool anyway and that is something that will handle the naming of those.

R/bm-array-to-vector.R
alistaire47: Ok, so my current thinking is that I want to close this PR and see if I can implement an opt-in case version via a function in the benchmark that takes in a case and appends a version to the params. The problem of how to version per benchmark is still there, but if we're not changing much, we can mostly ignore it.

I'm not sure yet whether the defaults written to the JSON (as a param, probably? negotiable) should be `1L`, `NULL`, or just not present; I might need to go look at conbench to see what would happen in each situation

jonkeane: Sounds good. Another (possible) approach would be to wait for this until benchmarks is no longer providing all the case info. On the arrow side we don't yet have an urgent need for version numbers in arrow. But I'm happy to do it in benchmarks for now and flesh out more uncertainty while we are thinking about it.

alistaire47: Yeah, I mostly want to make sure the code and logic paths are there to write JSON which can serve the purpose we need so we don't end up jumping through hoops to conform to what the endpoint needs

PR #105, opened 2022-06-08
Add capability to version cases

### LINKED ISSUE #101: Store a marker for benchmark versions

alistaire47:
Successor to #104; closes #101. This PR inserts versioning for cases into params via a benchmark-configurable function that takes a case (a named list of params) and returns an integer version.

There are a couple important questions here:

1. Should version be inserted as a param in the JSON (and therefore in conbench)? We're already inserting a lot of stuff into params that's arguably metadata (`cpu_count`, `lib_path`, a dataframe of `packages`); at some point we should move this stuff to a metadata section.
2. What should the default behavior be? This sets everything to version `1L`, but I haven't quite figured out yet whether this will break _all_ the histories in conbench. If so, it's probably better to change it to default to not setting it.

### DIFF
--- a/R/benchmark.R
+++ b/R/benchmark.R
@@ -82,6 +82,9 @@
 #' returning a `data.frame` of setup parameters. Use this to filter out invalid
 #' combinations of parameters (e.g. writing a Feather file with snappy
 #' compression, which is unsupported)
+#' @param case_version function taking a named list of setup parameters for a
+#' single case and returning an integer version for the case. Bump version to
+#' break conbench history for a case.
 #' @param ... additional attributes or functions, possibly called in `setup()`.
 #'
 #' @return A `Benchmark` object containing these functions
@@ -93,6 +96,7 @@ Benchmark <- function(name,
                       after_each = TRUE,
                       teardown = TRUE,
                       valid_params = function(params) params,
+                      case_version = function(params) 1L,
                       ...) {
   stopifnot(is.character(name))
   structure(
@@ -104,6 +108,7 @@ Benchmark <- function(name,
       after_each = substitute(after_each),
       teardown = substitute(teardown),
       valid_params = valid_params,
+      case_version = case_version,
       ...),
     class = "Benchmark"
   )
--- a/R/benchmark.R
+++ b/R/benchmark.R
@@ -84,7 +84,8 @@
 #' compression, which is unsupported)
 #' @param case_version function taking a named list of setup parameters for a
 #' single case and returning an integer version for the case, or `NULL` to not
-#' append a version. Changes to version will break conbench history for a case.
+#' append a version; `NA` will raise an error. Changes to version will break
+#' conbench history for a case.
 #' @param packages_used function taking a `data.frame` of setup parameters and
 #' returning a vector of R package names required
 #' @param ... additional attributes or functions, possibly called in `setup()`.
--- a/R/benchmark.R
+++ b/R/benchmark.R
@@ -83,8 +83,10 @@
 #' combinations of parameters (e.g. writing a Feather file with snappy
 #' compression, which is unsupported)
 #' @param case_version function taking a named list of setup parameters for a
-#' single case and returning an integer version for the case. Bump version to
-#' break conbench history for a case.
+#' single case and returning an integer version for the case, or `NULL` to not
+#' append a version. Changes to version will break conbench history for a case.
+#' @param packages_used function taking a `data.frame` of setup parameters and
+#' returning a vector of R package names required
 #' @param ... additional attributes or functions, possibly called in `setup()`.
 #'
 #' @return A `Benchmark` object containing these functions
@@ -96,7 +98,8 @@ Benchmark <- function(name,
                       after_each = TRUE,
                       teardown = TRUE,
                       valid_params = function(params) params,
-                      case_version = function(params) 1L,
+                      case_version = function(params) NULL,
+                      packages_used = function(params) "arrow",
                       ...) {
   stopifnot(is.character(name))
   structure(
@@ -109,6 +112,7 @@ Benchmark <- function(name,
       teardown = substitute(teardown),
       valid_params = valid_params,
       case_version = case_version,
+      packages_used = packages_used,
       ...),
     class = "Benchmark"
   )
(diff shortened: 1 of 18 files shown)


alistaire47:
Regarding 2, what exactly _does_ create a history discontinuity in conbench? @austin3dickey have you happened to learn this yet?

jonkeane:
> Regarding 2, what exactly does create a history discontinuity in conbench? @austin3dickey have you happened to learn this yet? 

We should [document this](https://github.com/conbench/conbench/issues/278) since it comes up frequently, but:

Differences in:
* case values "tags"
* contexts 
* hardware (though not _all_ values in a hardware are taken into account, the additional info, for example is allowed to vary)

https://github.com/conbench/conbench/blob/a3073041f702a4da8ad708afa776be6d7ab0ca72/conbench/entities/history.py#L35-L68

jonkeane:
> 1. Should version be inserted as a param in the JSON (and therefore in conbench)? We're already inserting a lot of stuff into params that's arguably metadata (cpu_count, lib_path, a dataframe of packages); at some point we should move this stuff to a metadata section.

> 2. What should the default behavior be? This sets everything to version 1L, but I haven't quite figured out yet whether this will break all the histories in conbench. If so, it's probably better to change it to default to not setting it.

We definitely shouldn't break history — we could either migrate and add 1s in everywhere or we could have a special case in conbench history where _for version_ `NULL == 1`

tests/testthat/test-run.R
```diff
@@ -41,7 +41,7 @@ test_that("get_params_summary returns a data.frame",{
 
   expected_summary <- dplyr::tibble(
     duration = 0.01, grid = TRUE, cpu_count = 1L,
-    output_type = "message", lib_path = "latest", did_error = FALSE
+    output_type = "message", case_version = 1L, lib_path = "latest", did_error = FALSE
```
jonkeane, line 44: It might be nice to have a test where you exercise this explicitly.

In this case, I'm less worried about testing the behavior itself and more having an example to follow of how it should look | work.

jonkeane:
> 1. Should version be inserted as a param in the JSON (and therefore in conbench)? We're already inserting a lot of stuff into params that's arguably metadata (cpu_count, lib_path, a dataframe of packages); at some point we should move this stuff to a metadata section.

Hmm, yeah I think this is ok. It is like many of the other variables that defines a case (at least on the conbench side)...

alistaire47:
> Differences in:
> 
>     * case values "tags"
> 
>     * contexts
> 
>     * hardware (though not _all_ values in a hardware are taken into account, the additional info, for example is allowed to vary)

Hm, so we are writing _most_ params (currently including `case_version` on this branch) to `tags`. I think the simplest approach is to default to not writing it then (but do write it when it's explicitly set!); otherwise we'll need special-casing elsewhere

tests/testthat/test-run.R
alistaire47: Yep, agreed. I want to clean this up a little and test for both the default case and the manually-set case

alistaire47: Added tests

R/benchmark.R
```diff
@@ -93,6 +98,8 @@ Benchmark <- function(name,
                       after_each = TRUE,
                       teardown = TRUE,
                       valid_params = function(params) params,
+                      case_version = function(params) NULL,
+                      packages_used = function(params) "arrow",
```
alistaire47, line 102: Added this because it's already called, but previously we were relying on every benchmark adding it in the same way though `...`. I'm now leaning more towards R6-ifying this class

boshek (approved):
Not a ton to say other than that it feels like until we actually need a version change we should stick with the `NULL == 1` situation.

alistaire47:
> Not a ton to say other than that it feels like until we actually need a version change we should stick with the `NULL == 1` situation.

Right now `NULL != 1`, because when it's `NULL` it won't add `case_version` to `tags`, but when it's `1` it will (and therefore will break history). Dunno if that's good, really, but it's easiest, as if we start defaulting everything to 1 it will break all the histories unless we go adjust everything

jonkeane (commented):
This is looking good — thanks! A few comments that might be quick to fix or a longer thread to unravel.

tests/testthat/test-run.R
```diff
@@ -34,6 +34,31 @@ test_that("run_one", {
...
+  res_unversioned <- run_benchmark(bm_unversioned)
+  lapply(res_unversioned$results, function(result) {
+    # when version is not supplied, it should not appear in tags
+    expect_false("case_version" %in% names(result$tags))
+  })
+
+  bm_versioned <- Benchmark(
+    "versioned",
+    setup = function(x = c('foo', 'bar')) cat(x),
+    case_version = function(params) c("foo" = 1L, "bar" = 2L)[params$x]
```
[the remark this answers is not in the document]
alistaire47: Because of how I constructed this example, if `params` were `list(x = "baz")` (or any novel value), it will return `NA`, which will get written to the tags as `"tags": {..., "case_version": null}`, which I _think_ may break history in conbench (though I'm not 💯).

R/run.R
```diff
@@ -186,6 +186,8 @@ run_bm <- function(bm, ..., n_iter = 1, profiling = FALSE, global_params = list(
   defaults <- lapply(get_default_args(bm$setup), head, 1)
   defaults$cpu_count <- parallel::detectCores()
   params <- modifyList(defaults, list(...))
+  params$case_version <- bm$case_version(params)
```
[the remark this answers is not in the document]
alistaire47: Yep, that's what I'm thinking; we should either do `stopifnot(is.na(case_version))` or automatically replace it with `NULL` (with a warning, probably?). I'm leaning towards erroring—we should see this when adjusting benchmarks (presuming we run them with all arg combinations), and we don't want to accidentally switch back to not versioning without noticing.

alistaire47:
ah crap my git branches got conflated! sorry, fixing

R/run.R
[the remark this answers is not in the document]
alistaire47: Ok, put in a `stopifnot()` and a test for it

PR #106, opened 2022-06-10
Benchmark for new taxi dataset

alistaire47:
Closes 96. Largely copies `bm-dataset-taxi-parquet.R`, but adjust queries for available fields and data structure. Also pushes `collect()`s to after `summarize()` as we can now (presumably we couldn't when the old one was written).

The new benchmark is pretty slow (~6m) because the dataset is CSVs; on parquet the queries run much faster. It might be interesting to extend `ensure_format()` so it works with datasets so we could compare the two, provided we don't mind the disk space usage.

The tests are wimpy because downloading the data takes a long time (~1h).


R/bm-dataset-taxi-2013.R
```diff
@@ -0,0 +1,106 @@
...
+    result <- NULL
+  },
+  run = {
+    result <- query$query(dataset)
+  },
+  after_each = {
+    query$assert(result)
+  },
+  cases = list(
+    vignette = list(
```
boshek, line 29: I wonder if this should be changed from "vignette" as that refers specifically to the vignette query.

boshek: Maybe "aggregation"?

R/bm-dataset-taxi-2013.R
```diff
@@ -0,0 +1,106 @@
...
+      },
+      assert = function(result) {
+        stopifnot(
+          identical(dim(result), c(5L, 3L)),
+          identical(names(result), c("payment_type", "tip_pct", "n")),
+          identical(sum(result$n), 4797187L)
+        )
+      }
+    ),
+    count_rows = list(
```
boshek, line 94: Ok this is pretty nit-picky but should this be called "count_dims"?

boshek (approved):
Nice! Looks good. Pretty straightforward. 

> Also pushes collect()s to after summarize() as we can now (presumably we couldn't when the old one was written).

Since you've identified this, would it be worth modifying the parquet taxi benchmark to reflect [this](https://github.com/apache/arrow/blob/53752adc6b81166cd4ee7db5a819494042f29197/r/vignettes/dataset.Rmd#L218-L228) new-ish query?

alistaire47:
> Since you've identified this, would it be worth modifying the parquet taxi benchmark to reflect [this](https://github.com/apache/arrow/blob/53752adc6b81166cd4ee7db5a819494042f29197/r/vignettes/dataset.Rmd#L218-L228) new-ish query?

We can, but I assume it will change the expected time, which is why I didn't. We could 

a. version it (though we can't pass that all the way through to conbench yet), or
b. call it something else (do we keep the old one then? I could go either way)

If we go with a., we should flip all the `collects()` in that one to last

R/bm-dataset-taxi-2013.R
alistaire47: Yeah, this is a little wrong. If we change the name on the old one (which this takes it from), it will break the history of that case, though; would we rather have them not be parallel or break history? Though if we want to flip all the collects in that one we might as well break all the history

R/bm-dataset-taxi-2013.R
alistaire47: yeah maybe i should just give up on trying to keep the names here parallel to taxi_parquet

boshek:
> a., we should flip all the collects() in that one to last

WRT to this specific point, I'd lean towards a. simply because I'm sure the relevance of a post `collect` summarise to arrow.

R/bm-dataset-taxi-2013.R
alistaire47: To be clear, I'm proposing changing this name on both, plus flipping all the `collect()`s in the other one to after the `summarize()`s.

Although because case versioning isn't getting picked up by {ursacomputing/benchmarks} yet history adding case versioning now won't actually break history till we implement it (technically just passing through `tags` from the JSON instead of {benchmarks} generating them itself). Is there another convenient way of breaking history? 

Or... 👿 do users care? My better angels are trying hard to stop me from doing the worse but much easier thing

jonkeane: I would vote for changing the names and breaking history now. I don't think a huge amount of focus is on these specific benchmarks, so if we suddenly have slightly different ones and the names are separate that's ok

jonkeane: And, FWIW we probably should have renamed this when we changed the benchmark to use `dim()` instead of `nrow()`

tests/testthat/test-bm-dataset-taxi-2013.R
```diff
@@ -0,0 +1,10 @@
+# Can't run benchmark because the data takes a long time to download (~1h), plus running is slow (~6m)
```
jonkeane, line 1: We don't _have to_ this PR, but we could make a small sample of this dataset that we [include in the package](https://github.com/ursacomputing/arrowbench/blob/018ce79135460b9f8b64e341f877f8a0e1f6433f/R/known-sources.R#L61-L79)

jonkeane (approved):
Looks good, one comment about how we might be able to test this

tests/testthat/test-bm-dataset-taxi-2013.R
alistaire47: Oh, I didn't realize those were part of the package! Let me see if I can get it to play nice with datasets as opposed to files without a lot of work, and if I can I'll add it.

R/bm-dataset-taxi-2013.R
alistaire47: Tweaked names and `collect()`s; lmk if y'all object to anything

tests/testthat/test-bm-dataset-taxi-2013.R
alistaire47: K, implemented `test_datasets` parallel to `test_sources` and added `taxi_2013_sample`, which is still 12 files but only 1000 rows

boshek: One request - can you include the code that you used to create the subset in a `data-raw` folder and then ignore that in the build?

alistaire47: Pushed!

PR #107, opened 2022-07-01
Harmonize Fannie Mae dataset and sample

### LINKED ISSUE #88: Harmonize Fannie Mae dataset and sample
Per https://github.com/ursacomputing/arrowbench/pull/87#discussion_r865188670 harmonize the full and sample Fannie Mae sources. Currently the sample has a bunch of null columns (it has 108 cols of which 61 have non-null data; the full dataset has 31 cols of which 29 are non-null) and neither has column names, so operations that rely on names generated from positions fail.

After this story, the only difference should be the number of rows. If we can find some real column names, that would be ideal.

alistaire47:
Closes #88 by adding an arrow schema for the Fannie Mae dataset and replacing the sample (which was not parallel; it was more true to the original than our dataset) with the first 100 lines. Documented sources for metadata.


R/known-sources.R
```diff
@@ -1,9 +1,52 @@
+# This is a reconstructed schema; it may not be completely accurate, but should
+# be serviceably so. For more information on variables, see the sources below
+# available from Fannie Mae. You may need to create an account first here:
+# https://datadynamics.fanniemae.com/data-dynamics/#/reportMenu;category=HP
+#
+# Data dictionary: https://capitalmarkets.fanniemae.com/resources/file/credit-risk/xls/crt-file-layout-and-glossary.xlsx
+# Names from R code here: https://capitalmarkets.fanniemae.com/media/document/zip/FNMA_SF_Loan_Performance_r_Primary.zip
+fannie_mae_schema <- arrow::schema(
+  LOAN_ID = arrow::string(),
+  ACT_PERIOD = arrow::string(),  # date. Monthly reporting period
```
jonkeane, line 10: ❤️ these comments. This is more a personal preference, but I wonder if they might lock _slightly_ better on the line above? Maybe following the line is the best option, but some of them get kind of long...

R/known-sources.R
```diff
@@ -1,9 +1,52 @@
...
+
+
 #' Known data files
 #' @export
 known_sources <- list(
   fanniemae_2016Q4 = list(
     url = "https://ursa-qa.s3.amazonaws.com/fanniemae_loanperf/2016Q4.csv.gz",
-    reader = function(file, ...) arrow::read_delim_arrow(file, delim = "|", col_names = FALSE, ...),
+    schema = fannie_mae_schema,
+    reader = function(file, ...) arrow::read_delim_arrow(file, delim = "|", schema = fannie_mae_schema, ...),
```
jonkeane, line 49: 🤔 Do we need to specify `fannie_mae_schema` twice here? Or is it possible to have the schema in `read_delim_arrow()` be populated from the first specification of `schema = fannie_mae_schema`.

It's not a huge deal to have it twice, but it might be nice to reuse it if it isn't too too much work

jonkeane (approved):
Looks good, two very very minor (possible) suggestions — neither required

R/known-sources.R
alistaire47: I didn't because it makes the list definition recursive, but because the function body doesn't get evaluated till it gets called, I think it's fine? I'll try it

R/known-sources.R
alistaire47: yeah black messed with some of these on the python side; I'll move them all

PR #108, opened 2022-07-08
Turn Fannie Mae schema into a function so schema instance gets created at runtime

alistaire47:
Bugfix cleaning up after commit to add Fannie Mae schema, which previous tried to maintain an instance of the class across sessions, which doesn't work and causes cryptic errors

### DIFF
--- a/R/known-sources.R
+++ b/R/known-sources.R
@@ -5,47 +5,49 @@
 #
 # Data dictionary: https://capitalmarkets.fanniemae.com/resources/file/credit-risk/xls/crt-file-layout-and-glossary.xlsx
 # Names from R code here: https://capitalmarkets.fanniemae.com/media/document/zip/FNMA_SF_Loan_Performance_r_Primary.zip
-fannie_mae_schema <- arrow::schema(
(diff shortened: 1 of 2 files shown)


R/known-sources.R
```diff
@@ -5,56 +5,58 @@
...
-  OTHER_FORECLOSURE_PROCEEDS = arrow::float64(),
-  NON_INTEREST_BEARING_UPB = arrow::float64(),
-  # all null
-  MI_CANCEL_FLAG = arrow::string(),
-  RE_PROCS_FLAG = arrow::string(),
-  # all null
-  LOAN_HOLDBACK_INDICATOR = arrow::string(),
-  SERV_IND = arrow::string()
-)
+fannie_mae_schema <- function() {
```
jonkeane, line 8:
```suggestion
fanniemae_schema <- function() {
```

for consistency?

jonkeane (approved):
Looks good, one small suggestion

R/known-sources.R
alistaire47: done

PR #111, opened 2022-08-19
Add datalogistik as datasource

jonkeane:


R/bm-tpc-h.R
```diff
@@ -107,8 +107,17 @@ tpc_h <- Benchmark("tpc_h",
...
+        # values in the answer(s) for those columns aren't found anywhere in the
+        # validation data or data generated with the official tool) so we try
+        # again with the duckdb specific answers
+        answer <- tpch_answer(scale_factor, query_id, datasource = "duckdb")
+        all_equal_out <- waldo::compare(result, answer, tolerance = 0.01)
+
+        if (length(all_equal_out) != 0) {
+          warning(paste0("\n", all_equal_out, "\n"))
+          stop("The answer does not match")
+        }
```
jonkeane, line 120: It should only really be necessary temporarily to support both duckdb's data and the real thing from datalogistik. But this will help us not need to coordinate anything like cache clearing for now.

R/ensure-tpch-source.R
```diff
@@ -4,36 +4,67 @@
...
+
+  if (exit_code != 0) {
+    warning(
+      'datalogistik not installed or on $PATH. If installed with pipx, ensure it ',
+      'is on $PATH, e.g. by adding PATH="${PATH}:${HOME}/.local/bin" to ~/.Renviron'
+    )
+  }
+
+  exit_code == 0
+}
```
jonkeane, line 18: I can work on something that does this if you're busy with other things, but ideally we would have a helper function that does the installation + helpful messages | docs if that doesn't work or `pipx` isn't installed, etc.

jonkeane:
Closed so that I could squash + rebase (but I wanted to keep some of this git history around and not squash it away...)

PR #112, opened 2022-09-13
export generate_tpch

boshek:
This is a minor PR to export the `generate_tpch` function.


R/ensure-tpch-source.R
```diff
@@ -4,6 +4,16 @@
...
+#'
+#' Generate tpch data at a given scale factor. By default,
+#' `data` is assumed to be relative to the current working directory, but
+#' you can set the environment variable `ARROWBENCH_DATA_DIR` to point to another
+#' (permanent) base directory.
+#'
+#' @param scale_factor a relative measure of the size of data in gigabytes.
+#'
+#' @export
 generate_tpch <- function(scale_factor = 1) {
```
alistaire47, line 17: Here's a branch with this via datalogistik: https://github.com/alistaire47/arrowbench/blob/feat/datalogistik/R/ensure-tpch-source.R If we're using this directly, we could just use datalogistik directly instead, though there is a nice util for parsing the json, so maybe this is nicer still.

R/ensure-tpch-source.R
```diff
@@ -4,6 +4,16 @@
 #' @export
 tpch_tables <- c("customer", "lineitem", "nation", "orders", "part", "partsupp", "region", "supplier")
 
+#' Generate tpch data
+#'
+#' Generate tpch data at a given scale factor. By default,
+#' `data` is assumed to be relative to the current working directory, but
+#' you can set the environment variable `ARROWBENCH_DATA_DIR` to point to another
+#' (permanent) base directory.
```
jonkeane, line 12: We should probably also mention something about the duckdb installation stuff that will ensue when running this, just so people don't freak out that there's compilations and stuff.

Also, what do you mean by "permanent" here?

jonkeane (approved):
I'm fine with this, a few comments about the docs. 

This isn't ideal (e.g. we could improve the way that duckdb installs now, _I think_), but we will soon move to a new thing, so let's not waste too much time on it.

R/ensure-tpch-source.R
boshek: > Also, what do you mean by "permanent" here?

It is from here: https://github.com/voltrondata-labs/arrowbench#known-data-sources-and-versions

To me it indicates that once you set it, all generated data with end up there. I will modify.

jonkeane: Huh, not a total surprise that I forgot that I (probably) wrote that or what I was intending at the time, but it strikes me as odd now 🙈. Thanks for removing it here (we could remove it form the readme too, though it's not an absolute requirement! that is bumping scope a bit)

jonkeane (approved):
Thanks, again for this!

PR #116, opened 2022-11-22
[DO NOT MERGE] Align results with those of labs/benchmarks and the Conbench API

alistaire47:
This PR changes a number of things, but all with the goal of aligning the JSON results arrowbench outputs with those currently posted to the Conbench API via voltrondata-labs/benchmarks so that eventually arrowbench can post its own results directly.

Changes fall into two categories:

1. Changes to result structure to align with the `/api/benchmarks/` endpoint
2. Changes to benchmarks so the data that populates results aligns with what labs/benchmarks generates. This has only been done for [the set of benchmarks run by arrow-benchmarks-ci](https://github.com/voltrondata-labs/arrow-benchmarks-ci/blob/0e4e2f4fb43adc22bf6f6c6beddbbb7881d557b5/config.py#L86-L91).

This PR cannot be merged until https://github.com/voltrondata-labs/benchmarks/pull/122 is merged, or it will break current arrow benchmarks. That PR contains code to handle changes to assorted names and parameter names here. It has been tested with that branch.

### DIFF
--- a/DESCRIPTION
+++ b/DESCRIPTION
@@ -1,7 +1,7 @@
 Package: arrowbench
 Type: Package
 Title: Tools for Continuous and Interactive Benchmarking
-Version: 0.1.0
+Version: 0.1.1
 Authors@R: c(
     person("Neal", "Richardson", email = "neal.p.richardson@gmail.com", role = "aut"),
     person("Jonathan", "Keane", email = "jkeane@gmail.com", role = c("aut", "cre"))
@@ -43,7 +43,7 @@ Suggests:
     RcppSimdJson,
     readr,
     vroom
-RoxygenNote: 7.1.2
+RoxygenNote: 7.2.1
 Roxygen: list(markdown = TRUE, load = "source")
 Collate: 
     'benchmark.R'
--- a/R/bm-df-to-table.R
+++ b/R/bm-df-to-table.R
@@ -6,8 +6,13 @@
 #' * `source` A known-file id to use (it will be read in to a data.frame first)
 #'
 #' @export
-df_to_table <- Benchmark("df_to_table",
-  setup = function(source = names(known_sources)) {
+df_to_table <- Benchmark("dataframe-to-table",
+  setup = function(source = c("chi_traffic_2020_Q1",
+                              "type_strings",
+                              "type_dict",
+                              "type_integers",
+                              "type_floats",
+                              "type_nested")) {
     source <- ensure_source(source)
     result_dim <- get_source_attr(source, "dim")
     # Make sure that we're not (accidentally) creating altrep vectors which will
--- a/R/bm-read-file.R
+++ b/R/bm-read-file.R
@@ -2,35 +2,35 @@
 #'
 #' @section Parameters:
 #' * `source` A known-file id, or a CSV(?) file path to read in
-#' * `format` One of `c("parquet", "feather", "fst")`
+#' * `file_type` One of `c("parquet", "feather", "fst")`
 #' * `compression` One of the values: `r paste(known_compressions, collapse = ", ")`
-#' * `output` One of `c("arrow_table", "data_frame")`
+#' * `output_type` One of `c("arrow_table", "data_frame")`
 #'
 #' @export
-read_file <- Benchmark("read_file",
-  setup = function(source = names(known_sources),
+read_file <- Benchmark("file-read",
+  setup = function(source = c("fanniemae_2016Q4", "nyctaxi_2010-01"),
                    # TODO: break out feather_v1 and feather_v2, feather_v2 only in >= 0.17
-                   format = c("parquet", "feather"),
+                   file_type = c("parquet", "feather"),
                    compression = c("uncompressed", "snappy", "lz4"),
-                   output = c("arrow_table", "data_frame")) {
-    # format defaults to parquet or feather, but can accept fst as well
-    format <- match.arg(format, c("parquet", "feather", "fst"))
-    # the output defaults are retrieved from the function definition (arrow_table and data_frame)
-    output <- match.arg(output)
+                   output_type = c("arrow_table", "data_frame")) {
+    # file_type defaults to parquet or feather, but can accept fst as well
+    file_type <- match.arg(file_type, c("parquet", "feather", "fst"))
+    # the output_type defaults are retrieved from the function definition (arrow_table and data_frame)
+    output_type <- match.arg(output_type)
 
     # ensure that we have the right kind of file available
-    input_file <- ensure_format(source, format, compression)
-    # retrieve the dimnesions for run-checking after the benchmark
+    input_file <- ensure_format(source, file_type, compression)
+    # retrieve the dimensions for run-checking after the benchmark
     result_dim <- get_source_attr(source, "dim")
 
-    # put the necesary variables into a BenchmarkEnvironment to be used when the
+    # put the necessary variables into a BenchmarkEnvironment to be used when the
     # benchmark is running.
     BenchEnvironment(
-      # get the correct read function for the input format
-      read_func = get_read_function(format),
+      # get the correct read function for the input file_type
+      read_func = get_read_function(file_type),
       input_file = input_file,
       result_dim = result_dim,
-      as_data_frame = output == "data_frame"
+      as_data_frame = output_type == "data_frame"
     )
   },
   # delete the results before each iteration
@@ -48,44 +48,44 @@ read_file <- Benchmark("read_file",
   },
   # validate that the parameters given are compatible
   valid_params = function(params) {
-    # make sure that the format and the compression is compatible
-    # and fst doesn't have arrow_table output
-    drop <- !validate_format(params$format, params$compression) |
-            params$output == "arrow_table" & params$format == "fst"
+    # make sure that the file_type and the compression is compatible
+    # and fst doesn't have arrow_table output_type
+    drop <- !validate_format(params$file_type, params$compression) |
+            params$output_type == "arrow_table" & params$file_type == "fst"
     params[!drop,]
   },
-  # packages used when specific formats are used
+  # packages used when specific file_types are used
   packages_used = function(params) {
     pkg_map <- c(
       "feather" = "arrow",
       "parquet" = "arrow",
       "fst" = "fst"
     )
-    pkg_map[params$format]
+    pkg_map[params$file_type]
   }
 )
 
 #' Get a reader
 #'
-#' @param format what format to read
+#' @param file_type what file_type to read
 #'
 #' @return the read function to use
 #' @export
-get_read_function <- function(format) {
+get_read_function <- function(file_type) {
   pkg_map <- c(
     "feather" = "arrow",
     "parquet" = "arrow",
     "fst" = "fst"
   )
-  library(pkg_map[[format]], character.only = TRUE, warn.conflicts = FALSE)
+  library(pkg_map[[file_type]], character.only = TRUE, warn.conflicts = FALSE)
 
-  if (format == "feather") {
+  if (file_type == "feather") {
     return(function(...) arrow::read_feather(...))
-  } else if (format == "parquet") {
+  } else if (file_type == "parquet") {
     return(function(...) arrow::read_parquet(...))
-  } else if (format == "fst") {
+  } else if (file_type == "fst") {
     return(function(..., as_data_frame) fst::read_fst(...))
   } else {
-    stop("Unsupported format: ", format, call. = FALSE)
+    stop("Unsupported file_type: ", file_type, call. = FALSE)
   }
 }
--- a/R/bm-tpc-h.R
+++ b/R/bm-tpc-h.R
@@ -14,7 +14,7 @@
 #'
 #' @importFrom waldo compare
 #' @export
-tpc_h <- Benchmark("tpc_h",
+tpc_h <- Benchmark("tpch",
   setup = function(engine = "arrow",
                    query_id = 1:22,
                    format = c("native", "parquet"),
@@ -134,6 +134,14 @@ tpc_h <- Benchmark("tpc_h",
       # TODO: do this?
       ( params$engine == "dplyr" & params$format == "native" )
     params[!drop,]
+  },
+  batch_id_fun = function(params) {
+    batch_id <- uuid()
+    paste0(batch_id, "-", params$scale_factor, substr(params$format, 1, 1))
+  },
+  tags_fun = function(params) {
+    params$query_id <- sprintf("TPCH-%02d", params$query_id)
+    params
   },
   # packages used when specific formats are used
   packages_used = function(params) {
--- a/R/result.R
+++ b/R/result.R
@@ -102,6 +102,10 @@ Serializable <- R6Point1Class(
         if (inherits(element, "Serializable")) {
           element <- element$list
         }
+        # iterate through lists of serializables and recurse
+        if (is.list(element) && all(vapply(element, function(el) inherits(el, "Serializable"), logical(1)))) {
+          element <- lapply(element, function(el) el$list)
+        }
         element
       })
     },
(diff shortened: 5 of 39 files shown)


alistaire47:
Failing build is a DuckDB thing that I believe will be addressed by https://github.com/voltrondata-labs/arrowbench/pull/113

DESCRIPTION
```diff
@@ -1,7 +1,7 @@
 Package: arrowbench
 Type: Package
 Title: Tools for Continuous and Interactive Benchmarking
-Version: 0.1.0
+Version: 0.1.1
```
boshek, line 4: Feels to me that this should increment a little her. 0.2.0

DESCRIPTION
```diff
@@ -1,7 +1,7 @@
...
 Authors@R: c(
     person("Neal", "Richardson", email = "neal.p.richardson@gmail.com", role = "aut"),
     person("Jonathan", "Keane", email = "jkeane@gmail.com", role = c("aut", "cre"))
```
boshek, line 7: To me, this is a substantial change and you should probably add your name to this list.

boshek (approved):
😍 this is amazing! What a great effort to standardize. No substantive changes though I think adding your name to the maintainer list is a good idea. 

I also wonder with a version change if a NEWS file is merited here. I think an item for this version can still be brief and high-level. Stealing from your comment, something like:

> This version of arrowbench aligns the JSON results arrowbench outputs with those currently posted to the Conbench API via voltrondata-labs/benchmarks so that eventually arrowbench can post its own results directly to a conbench instance

R/bm-read-file.R
```diff
@@ -2,35 +2,35 @@
...
+                   file_type = c("parquet", "feather"),
                    compression = c("uncompressed", "snappy", "lz4"),
-                   output = c("arrow_table", "data_frame")) {
-    # format defaults to parquet or feather, but can accept fst as well
-    format <- match.arg(format, c("parquet", "feather", "fst"))
-    # the output defaults are retrieved from the function definition (arrow_table and data_frame)
-    output <- match.arg(output)
+                   output_type = c("arrow_table", "data_frame")) {
+    # file_type defaults to parquet or feather, but can accept fst as well
+    file_type <- match.arg(file_type, c("parquet", "feather", "fst"))
```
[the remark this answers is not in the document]
alistaire47: We can still run it for comparison if we want, though AFAIK we don't have a policy on when we actually do run external packages and compare against them. It might be nice to run these when there's an Arrow release and include them for context in that report? Not priority, but maybe useful context.

Or was {fst} deprecated or something? I thought I remembered something like that, but now I don't see it.

R/result.R
```diff
@@ -163,124 +172,109 @@ BenchmarkResult <- R6Point1Class(
...
+      d <- self$optional_benchmark_info$params
 
       d$packages <- NULL
 
       to_list_col <- lengths(d) != 1
       d[to_list_col] <- lapply(d[to_list_col], list)
 
       d <- dplyr::as_tibble(d)
-      d$did_error <- FALSE
+      d$did_error <- !is.null(self$error)
```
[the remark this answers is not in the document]
alistaire47: Error handling is an area where we can go further in the future if there's bandwidth and there are enough flaky benchmarks. With the new `validation` key on `/api/benchmarks/`, we should really be reporting failure to produce the right thing separately from failure to run properly. And in theory we could populate the `error_info` and `error_type` on `PUT /api/runs/{run_id}/` too when setup or teardown fails (and not running or validation).

But in practice most of these don't fail often enough to make the effort very worthwhile. Which is good!

R/bm-read-file.R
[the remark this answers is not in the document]
boshek: I think maybe you are thinking [disk.frame](https://cran.r-project.org/web/packages/disk.frame/readme/README.html). And I think the subtext here is that it really doesn't matter too much to include it so why remove. Convinced.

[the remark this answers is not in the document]
alistaire47: Oh, yep, that was it! Thanks

R/bm-tpc-h.R
```diff
@@ -14,7 +14,7 @@
 #'
 #' @importFrom waldo compare
 #' @export
-tpc_h <- Benchmark("tpc_h",
+tpc_h <- Benchmark("tpch",
```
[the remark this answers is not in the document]
boshek: > These name changes are a pain

100%. Thank you for doing them! 
I'll leave it up to you. It is not a material difference and really just reflects a slight nit.

alistaire47:
arrow-benchmarks-ci buildkite test build (together with voltrondata-labs/benchmarks#122) passes: https://buildkite.com/apache-arrow/arrow-bci-benchmark-on-arm64-m6g-linux-compute/builds/1740#0184ca79-7c13-407e-ad17-814eed712a28/6-14249

.github/workflows/R-CMD-check.yaml
```diff
@@ -1,4 +1,8 @@
-on: [push, pull_request]
+on:
+  push:
+    branches:
+      - main
+  pull_request:
```
jonkeane, line 5: Any reason for only running on push to main? I definitely from time to time appreciate having CI on branches even before they get to the PR stage

DESCRIPTION
```diff
@@ -1,10 +1,11 @@
 Package: arrowbench
 Type: Package
 Title: Tools for Continuous and Interactive Benchmarking
-Version: 0.1.0
+Version: 0.2.0
 Authors@R: c(
     person("Neal", "Richardson", email = "neal.p.richardson@gmail.com", role = "aut"),
-    person("Jonathan", "Keane", email = "jkeane@gmail.com", role = c("aut", "cre"))
+    person("Jonathan", "Keane", email = "jkeane@gmail.com", role = c("aut", "cre")),
+    person("Edward", "Visel", email = "edward.visel@gmail.com", role = "aut", comment = c(ORCID = "0000-0002-2811-6254"))
```
jonkeane, line 8: 🎉

R/bm-df-to-table.R
```diff
@@ -6,8 +6,13 @@
...
 #' @export
-df_to_table <- Benchmark("df_to_table",
-  setup = function(source = names(known_sources)) {
+df_to_table <- Benchmark("dataframe-to-table",
+  setup = function(source = c("chi_traffic_2020_Q1",
+                              "type_strings",
+                              "type_dict",
+                              "type_integers",
+                              "type_floats",
+                              "type_nested")) {
```
jonkeane, line 15: Very minor style nit, though this could in principle go both ways I find the following _slightly_ easier to read:

```suggestion
  setup = function(source = c(
    "chi_traffic_2020_Q1",
    "type_strings",
    "type_dict",
    "type_integers",
    "type_floats",
    "type_nested"
    )) {
```

R/bm-read-file.R
```diff
@@ -2,35 +2,35 @@
 #'
 #' @section Parameters:
 #' * `source` A known-file id, or a CSV(?) file path to read in
-#' * `format` One of `c("parquet", "feather", "fst")`
+#' * `file_type` One of `c("parquet", "feather", "fst")`
 #' * `compression` One of the values: `r paste(known_compressions, collapse = ", ")`
-#' * `output` One of `c("arrow_table", "data_frame")`
+#' * `output_type` One of `c("arrow_table", "data_frame")`
```
jonkeane, line 7: I'm fine changing these here if it's easiest to go this direction, though I kinda wish we went the other way (but understand migrating out of that would be a pain in the ass...)

R/bm-read-file.R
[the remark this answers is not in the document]
jonkeane: Yeah, we might need to drop support for fst when we move to datalogistik (or do something slightly different if we were to use it), but for now let's keep it here

R/bm-tpc-h.R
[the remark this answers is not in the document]
jonkeane: This is another example where I wish we had called it `tpc-h` or `tpc_h` in what was stored in conbench in the first place. Let's go with `tpch` for now to reduce the number of migration bits we need, but another good example of how the spirit of this project of getting arrowbench -> conbench more directly will help make sure we put up what we want and not what another shim layer between had

R/bm-tpc-h.R
```diff
@@ -135,6 +135,18 @@ tpc_h <- Benchmark("tpc_h",
       ( params$engine == "dplyr" & params$format == "native" )
     params[!drop,]
   },
+  batch_id_fun = function(params) {
+    batch_id <- uuid()
+    paste0(batch_id, "-", params$scale_factor, substr(params$format, 1, 1))
```
jonkeane, line 140: I know this is effectively copy|paste from voltrondata-labs/benchmarks but ouch that's kinda gross. Would you mind adding a todo here to make this be a bit nicer (probably need to properly stringify the scale factor and also have the format be full(er)?)

R/result.R
```diff
@@ -163,124 +172,109 @@ BenchmarkResult <- R6Point1Class(
...
     tags = function(tags) private$get_or_set_serializable(variable = "tags", value = tags),
     info = function(info) private$get_or_set_serializable(variable = "info", value = info),
+    optional_benchmark_info = function(optional_benchmark_info) private$get_or_set_serializable(variable = "optional_benchmark_info", value = optional_benchmark_info),
+    machine_info = function(machine_info) private$get_or_set_serializable(variable = "machine_info", value = machine_info),
+    cluster_info = function(cluster_info) private$get_or_set_serializable(variable = "cluster_info", value = cluster_info),
     context = function(context) private$get_or_set_serializable(variable = "context", value = context),
     github = function(github) private$get_or_set_serializable(variable = "github", value = github),
-    options = function(options) private$get_or_set_serializable(variable = "options", value = options),
-    output = function(output) private$get_or_set_serializable(variable = "output", value = output),
-    rscript = function(rscript) private$get_or_set_serializable(variable = "rscript", value = rscript),
```
jonkeane, line 229: I know that we can't send it yet, but it feels a little silly to pull this out when we will then put it back in (ideally) very soon. Anyway that we can not delete this just to re-add it later?

R/run.R
```diff
@@ -200,20 +216,39 @@ run_bm <- function(bm, ..., n_iter = 1, profiling = FALSE, global_params = list(
...
 
   out <- BenchmarkResult$new(
-    name = bm$name,
-    result = do.call(rbind, results),
-    params = all_params,
+    run_name = NULL,
+    run_id = NULL,
+    batch_id = batch_id,
+    run_reason = NULL,
+    # timestamp = utc_now_iso_format(),  # let default populate
```
jonkeane, line 229: Do we need this comment?

jonkeane (commented):
Thanks for this, it's looking good. I don't have too many substantive comments

jonkeane:
Oh, and also that release check is actually fixed by https://github.com/voltrondata-labs/arrowbench/pull/117/files#diff-e35a10b8001b4a29a3a67979b010eebbf6285e9b0369c78693662e52d7b9e3b3R12 it's ultimately not going to be needed once we move to datalogistik, but it'll be fixed in main when #117 is merged

.github/workflows/R-CMD-check.yaml
alistaire47: We were running duplicate builds of everything on PRs because they're both branches and pull requests, so I copied this from one of our other repos. We could drop lines 4-5 (I think?) if we want builds on all pushes to branches

R/bm-df-to-table.R
alistaire47: Yeah, that's black style, kind of. It might look weird if there were more than one parameter, though; are we ok with 

```r
  setup = function(
    source = c(
      "chi_traffic_2020_Q1",
      "type_strings",
      "type_dict",
      "type_integers",
      "type_floats",
      "type_nested"
    )
  ) {
    source <- ensure_source(source)
```
?

R/bm-read-file.R
alistaire47: Yeah, I definitely don't love these changes, but they're the easiest way to keep consistency with history. There's nothing stopping us from changing back and migrating the database in the future if we want, though.

R/bm-tpc-h.R
alistaire47: Sure. We could also change it now, if you like; the only thing I can think of that might (? not sure) care about consistency here is @boshek's release report.

R/result.R
alistaire47: Per one of our earlier conversations, they're removed from the top-level fields here so the keys exactly correspond to the `/api/benchmarks/` endpoint, but we're not losing data, because we're sticking it all in `optional_benchmark_info` starting here: https://github.com/voltrondata-labs/arrowbench/pull/116/files/1019e29cc2053afc143943b6bf5094c50af0e903#diff-c118b736cd58f32580f07445199c10b7fec49b44f25f27cd46bbf1215cc9eff3R241-R247 (and `output` and `rscript` are appended later). They're not part of the schema anymore, though.

For a bit I tried to make a slim result object that was just the API schema and a richer one that had extra metadata like this that inherited from it, but it got pretty complicated and ugly, to be honest. This way is less structured, but what structure we have is all consistent such that we don't have 6 versions of results schemas like before, which is nice.

R/run.R
alistaire47: Not really, I just left it so you can see all the fields on the object here even if they're not getting filled in here. It's a little confusing to put it at the end of the line as it's really why the line is commented out; I could put it as a line before, or just drop it all. Or we could actually specify timestamp here. It's all minor.

.github/workflows/R-CMD-check.yaml
jonkeane: > We were running duplicate builds of everything on PRs because they're both branches and pull requests

Could you point me to an example of this?

R/bm-df-to-table.R
jonkeane: Yup, that's fine too. Mostly I try to avoid the more-indented-than-the-block-should-be-otherwise when I can (though I know it makes sense for function definitions and why RStudio does what it does automatically here...)

R/bm-read-file.R
jonkeane: Yeah, and if we do decided to do it in the future, doing it separate from these specific changes will make it easier to both do and to diagnose if something goes wrong

.github/workflows/R-CMD-check.yaml
alistaire47: On [your PR](https://github.com/voltrondata-labs/arrowbench/pull/119), if you expand the builds, there are two versions of each, one appended with `(pull_request)`, the other with `(push)`: 

image

R/result.R
jonkeane: Aaah right, sticking them in `optional_benchmark_info` is great! I didn't see that directly here, but now see it  in `run.R` Thanks!

.github/workflows/R-CMD-check.yaml
jonkeane: IIUC, those are testing two different things though (check out the details of the checkouts). the `push` is the `HEAD` of the branch, the `pull_request` is the merge of `HEAD` of branch + main. For _that specific_ PR they end up being identical code wise (since I rebased the branch on main a few moments ago), but they can and frequently are quite different

## push

```
Checking out the ref
  /usr/local/bin/git checkout --progress --force -B duckdb-ext refs/remotes/origin/duckdb-ext
  Switched to a new branch 'duckdb-ext'
  branch 'duckdb-ext' set up to track 'origin/duckdb-ext'.
/usr/local/bin/git log -1 --format='%H'
'14f9a4c5797ef2529328c050e512e18010737531'
```
https://github.com/voltrondata-labs/arrowbench/actions/runs/3605143011/jobs/6075232144#step:2:455

## pull_request

```
Checking out the ref
  /usr/local/bin/git checkout --progress --force refs/remotes/pull/119/merge
  HEAD is now at 1dfb8c9 Merge 14f9a4c5797ef2529328c050e512e18010737531 into ada600dea8d6383b802be893b9e26b6b6b38792a
/usr/local/bin/git log -1 --format='%H'
'1dfb8c955aeb476fdc7e6dd055db57cacaa3f2b7'
```
https://github.com/voltrondata-labs/arrowbench/actions/runs/3605143166/jobs/6075232423#step:2:459

alistaire47: Hm, interesting. Though it seems like `pull_request` is what I'd expect aside from on the default branch. But I can revert.

R/bm-tpc-h.R
alistaire47: Put in a `TODO` for now

jonkeane (approved):
This is looking great. Thank you!

PR #118, opened 2022-12-02
Bump GHA versions

jonkeane:
The proximate cause was the coverage test not setting up correctly, but bumped all the r-lib actions to v2, and checkout to v3


alistaire47:
Will this fix issues like this: https://github.com/voltrondata-labs/arrowbench/actions/runs/3603731943/jobs/6072283074#step:3:63 Just cropped up when trying to rebase

jonkeane:
Yup, that's exactly the thing that made me push on this (and it does resolve that 🎉 )

PR #119, opened 2022-12-02
Use the duckdb extension instead of (re)compiling

jonkeane:
With the DuckDB 0.6.0 release, we can (reliably) use the extension method. This means we don't need to re-compile duckdb from source (finally). This code will be less important when we move to datalogistik (though we might still use it for answers in some circumstances.

But most importantly it shrinks the time for CI to run from 1.5–2 hours down to 6 minutes for the relevant job.


alistaire47 (approved):

boshek (approved):
LGTM!

PR #122, opened 2022-12-12
Add a top-level runner for sets of benchmarks

alistaire47:
This PR adds a top-level runner for managing and running sets of benchmarks at once, an important step in moving towards arrow-benchmarks-ci calling arrowbench directly instead of through voltrondata-labs/benchmarks.

The key addition of this PR is a S3-classed dataframe called `BenchmarkDataFrame` that contains at least

- benchmark `name` (the attribute that will be the `name` tag, not the object name)
- the benchmark object itself in a list column
- a list column of parameter dataframes (which can be `NULL`, in which case they'll be populated with `default_params()`)

After they're run, a `results` column will be appended, which is a list column where each element is an instance of `BenchmarkResults` containing an instance of `BenchmarkResult` for each case (row) in the parameter dataframe.

Because it's just a dataframe (and a tibble, mostly for nice list-column printing), users can filter how they like, modify the `parameters` column however they see fit prior to running, and expand and parse results at their leisure after running, and everything will be kept associated properly because everything related is in a row.

Thus far (as of writing; this may change by the time this gets merged), I've resisted the urge to add a bunch of helper functions to, say, unnest parameters and results in parallel or some other version of `get_params_summary()` or whatever. I'm open to thoughts on what's useful here; there's a lot that could be added, but quite a lot is already possible with a line of code or two.

This PR also adds an exported vector named `URSA_I9_9960X_R_BENCHMARK_NAMES`, which corresponds exactly with [this list in arrow-benchmarks-ci](https://github.com/voltrondata-labs/arrow-benchmarks-ci/blob/main/config.py#L85-L92) of the list of R benchmarks run regularly. (Several machines also run `tpch` alone.) I'm not sure this is the best way to manage these, but I'm very open to suggestions; I know we do want to manage this list from arrowbench, but we haven't thus far said exactly how. We could stick it in a list as a config or make a little function that will return a set of benchmark names for a machine name or something? There are options.

This PR does not yet add code for posting results to Conbench; that will be in the next PR. You can see where it will fit, though; we could post results as they're completed with `run()` or a finished set of results from a `BenchmarkDataFrame`.


R/benchmark-dataframe.R
```diff
@@ -0,0 +1,113 @@
...
+#' @return An instance of [BenchmarkDataFrame] with all the benchmarks contained
+#' by a package
+#'
+#' @export
+get_package_benchmarks <- function(package = "arrowbench") {
+  nms <- getNamespaceExports(package)
+  objs <- mget(nms, envir = getNamespace(package))
+  bms <- Filter(function(x) inherits(x, "Benchmark"), objs)
+  BenchmarkDataFrame(benchmarks = bms)
+}
```
alistaire47, line 64: This is some introspective R dark magic, but because these lines can only run after this package is installed, I'm fairly confident they're safe, and they keep us from needing to maintain a list of available benchmarks, which I like a lot.

R/benchmark.R
```diff
@@ -136,10 +136,33 @@ Benchmark <- function(name,
...
+#' @param ... Named arguments corresponding to the parameters of `bm`'s `setup`
+#' function. May also contain `cpu_count`, `lib_path`, and `mem_alloc`.
+#'
+#' @return For `default_params.Benchmark`, a dataframe of parameter combinations
+#' to try with a column for each parameter and a row for each combination.
+#'
+#' @export
+default_params <- function(x, ...) {
+  UseMethod("default_params")
+}
```
alistaire47, line 154: Since this will be exported, do we want to rename this to `get_default_parameters()` or something? We don't need to; it's just a convenient time to do so if we want

DESCRIPTION
```diff
@@ -15,8 +15,6 @@ Depends: R (>= 3.5.0)
 Imports:
     arrow,
     bench,
-    callr,
-    dbplyr,
```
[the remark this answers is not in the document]
alistaire47: Ok by the builds, I think we do need dbplyr in Suggests because of the tpch tests

R/benchmark-dataframe.R
austin3dickey: this is cool! So is there no way to override the default_params when you use this function?

R/benchmark.R
austin3dickey: Yes, IMO

tests/testthat/test-benchmark-dataframe.R
```diff
@@ -0,0 +1,140 @@
...
+  assert_benchmark_dataframe(bm_df_augmented, bm_list, lapply(bm_list, default_params, duration = 1))
+  lapply(bm_df_augmented$parameters, function(param_df) {
+    expect_s3_class(param_df, "data.frame")
+    expect_equal(param_df, default_params(placebo, duration = 1))
+    expect_gt(nrow(param_df), 0L)
+  })
+})
+
+
+test_that("run() dispatches and run.default() errors", {
```
austin3dickey, line 87: Nice coverage as always

R/benchmark-dataframe.R
```diff
@@ -0,0 +1,113 @@
+#' A vector of benchmark attribute names run on `ursa-i9-9960x`
+#'
+#' @export
+URSA_I9_9960X_R_BENCHMARK_NAMES <- c(
```
austin3dickey, line 4: I'm fine with this config looking like this for now (we could turn it into a function once you know more use cases) but maybe it should be in its own file?

austin3dickey (approved):
Looks great!

R/benchmark-dataframe.R
alistaire47: Good question, and no, I went with making it explicit by defining `default_params.BenchmarkDataFrame()` so you can do 

```r
get_package_benchmarks() |>
    default_params(cpu_count = arrow::cpu_count()) |>  # or whatever
    run()
```

Or you can put kwargs directly in `run()` and let the params dataframe be generated on the fly, but from a workflow perspective it was definitely useful to be able to see the params before running anything.

The big caveat here is that this is only really useful for params relevant to all the benchmarks in the dataframe, i.e. things like `cpu_count`. So it becomes way more important that our defaults are right (which I did a lot of work in the last PR to correct at least for the ones we're running regularly). For anything else, you'll probably need to assign to a subset with

```r
bm_df <- get_package_benchmarks()
bm_df$parameters[bm_df$name == "read_json"] <- default_params(read_json, source = c("fanniemae_2016", "nyctaxi_2010"))
```
or whatever. I don't love this, but don't see a better alternative, really.

R/benchmark-dataframe.R
alistaire47: Yeah, I'm thinking a few things in this file should move, really; all the `run()` stuff should go in `run.R`; `default_params.BenchmarkDataFrame()` should move to be with the rest of the methods (which should probably be split into their own file), and this should go in `config.R` or something, probably. So TODOs:

- [ ] move `run()`
- [ ] move `default_params.BenchmarkDataFrame()`
- [ ] move config

alistaire47: done

R/benchmark.R
alistaire47: done, felt like time.

R/params.R
```diff
@@ -0,0 +1,66 @@
...
+#'
+#' @return For `get_default_parameters.Benchmark`, a dataframe of parameter combinations
+#' to try with a column for each parameter and a row for each combination.
+#'
+#' @export
+get_default_parameters <- function(x, ...) {
+  UseMethod("get_default_parameters")
+}
+
+#' @export
```
jonkeane, line 18: Do we want or need `#' @rdname get_default_parameters` here?

jonkeane (approved):
This is looking great, thanks for all the work here!

One thing I noticed in your description is the difference between these two groups goes to show just how much we're leaving on the table by not running benchmarks we've already crafted, but "just" hadn't wired up through labs/benchmarks yet:

```
> get_package_benchmarks()
# <BenchmarkDataFrame>
# A tibble: 14 × 3
   name                         benchmark    parameters
 * <chr>                        <named list> <list>    
 1 dataset_taxi_2013            <Benchmrk>   <NULL>    
 2 row_group_size               <Benchmrk>   <NULL>    
 3 read_csv                     <Benchmrk>   <NULL>    
 4 write_csv                    <Benchmrk>   <NULL>    
 5 read_json                    <Benchmrk>   <NULL>    
 6 remote_dataset               <Benchmrk>   <NULL>    
 7 file-write                   <Benchmrk>   <NULL>    
 8 dataframe-to-table           <Benchmrk>   <NULL>    
 9 table_to_df                  <Benchmrk>   <NULL>    
10 array_to_vector              <Benchmrk>   <NULL>    
11 partitioned-dataset-filter   <Benchmrk>   <NULL>    
12 file-read                    <Benchmrk>   <NULL>    
13 tpch                         <Benchmrk>   <NULL>    
14 array_altrep_materialization <Benchmrk>   <NULL>  

> get_package_benchmarks() %>% filter(name %in% URSA_I9_9960X_R_BENCHMARK_NAMES)
# <BenchmarkDataFrame>
# A tibble: 5 × 3
  name                       benchmark    parameters
  <chr>                      <named list> <list>    
1 file-write                 <Benchmrk>   <NULL>    
2 dataframe-to-table         <Benchmrk>   <NULL>    
3 partitioned-dataset-filter <Benchmrk>   <NULL>    
4 file-read                  <Benchmrk>   <NULL>    
5 tpch                       <Benchmrk>   <NULL>  
```

I'll be very glad to see the day when adding benchmarks to what's run is only a matter of determining if we have enough time to run them under load as opposed to that + the work to wire them all up into labs/benchmarks too.

Am I interpreting correctly that effectively what we would need to add to arrow-benchmarks-ci (after we add the publishing, of course) is something like:

```
# prior to executing this r script, ensure that arrowbench is installed + up to date + the arrow R package is too
library(arrowbench)

get_package_benchmarks() %>% filter(name %in% URSA_I9_9960X_R_BENCHMARK_NAMES) %>% run()
```

Is that right? If so, I wonder if it might make sense to have the list `URSA_I9_9960X_R_BENCHMARK_NAMES` be explicit and in the arrow-benchmarks-ci repo itself (e.g. `get_package_benchmarks() %>% filter(name %in%c("dataframe-to-table", "df_to_table", "file-read", "file-write", "partitioned-dataset-filter", "tpch")) %>% run()`)? Is there any other reason we might need the list of benchmarks scheduled on the ursa i9 to live inside of arrowbench?

> This PR does not yet add code for posting results to Conbench; that will be in the next PR. You can see where it will fit, though; we could post results as they're completed with run() or a finished set of results from a BenchmarkDataFrame.

I assume you'll also add in [the cache dropping](https://github.com/conbench/conbench/blob/508267f8159f0b932f81b942eb5fe0e5dccf4869/conbench/runner.py#L306-L313) in the follow on PR as well so that we have similar results across the two runners? For now I would "just" copy that code over and implement it in arrowbench. Ideally it's part of some dependency-lite utility set that we could depend on — but I don't think we've yet to find a way to truly have an easy way to do that and it's 1(ish) bash command that copying around is the least-painful way to get to where we need to go.

R/benchmark-dataframe.R
boshek: This is really nice and nice that it works with `load_all()`. Sometimes that can be a little hairy.

boshek (approved):
This looks amazing. I do wonder if it might worthwhile having an example (beyond what is in the tests) in the README or something.

R/params.R
alistaire47: Nothing complains if you don't, but it does make it easier to see what classes have methods, so I'll add it

R/benchmark.R
```diff
@@ -129,44 +129,14 @@ Benchmark <- function(name,
   )
 }
 
+
```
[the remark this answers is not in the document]
alistaire47: I think I've got `black` formatting so ingrained in my brain now that I default to two newlines between top-level classes/functions in a file. I kind of like it, though; it makes it easier to see the breaks when scanning quickly.

alistaire47:
> Am I interpreting correctly that effectively what we would need to add to arrow-benchmarks-ci (after we add the publishing, of course) is something like:
> 
> ```
> # prior to executing this r script, ensure that arrowbench is installed + up to date + the arrow R package is too
> library(arrowbench)
> 
> get_package_benchmarks() %>% filter(name %in% URSA_I9_9960X_R_BENCHMARK_NAMES) %>% run()
> ```
> 
> Is that right? If so, I wonder if it might make sense to have the list `URSA_I9_9960X_R_BENCHMARK_NAMES` be explicit and in the arrow-benchmarks-ci repo itself (e.g. `get_package_benchmarks() %>% filter(name %in%c("dataframe-to-table", "df_to_table", "file-read", "file-write", "partitioned-dataset-filter", "tpch")) %>% run()`)? Is there any other reason we might need the list of benchmarks scheduled on the ursa i9 to live inside of arrowbench?

Yep, that's roughly the intended pattern. I thought about adding a CLI or some other sort of simplified entrypoint, but this seems simple enough it's not really worth the effort.

As to where that vector lives, yes, that's the part of this PR I'm least certain of. I thought we wanted to move that out of arrow-benchmarks-ci into arrowbench, but the more I look at it, the more I think it should stay where it is, and we just cut out labs/benchmarks in the middle. It may be useful to keep this around for testing, but I may at least un-export it before merging if I don't delete it entirely.

> > This PR does not yet add code for posting results to Conbench; that will be in the next PR. You can see where it will fit, though; we could post results as they're completed with run() or a finished set of results from a BenchmarkDataFrame.
> 
> I assume you'll also add in [the cache dropping](https://github.com/conbench/conbench/blob/508267f8159f0b932f81b942eb5fe0e5dccf4869/conbench/runner.py#L306-L313) in the follow on PR as well so that we have similar results across the two runners? For now I would "just" copy that code over and implement it in arrowbench. Ideally it's part of some dependency-lite utility set that we could depend on — but I don't think we've yet to find a way to truly have an easy way to do that and it's 1(ish) bash command that copying around is the least-painful way to get to where we need to go.

Ah right. I'm of multiple minds here; we could insert it into benchconnect, but a. it doesn't quite fit with the rest of what's there, and b. I think we want to call that as close to running as possible, and wrapping it in a CLI pushes it further. And wrapping it in a CLI still wouldn't standardize whether it's called per iteration, or before a separate warmup iteration, or once per run anyway—that would still depend on the runner. So yeah, I guess the least-bad is to copy this all over. Created #126 to do it; it should probably be its own little tiny PR, not part of the benchconnect mess, as it's unrelated.

> I do wonder if it might worthwhile having an example (beyond what is in the tests) in the README or something.

Yeah, if this is going to be the way to run a bunch of benchmarks, it should be in the README, if nothing else so it's clear how to run benchmarks in CI without looking at arrow-benchmarks-ci. I'll add a bit.

jonkeane:
All of that sounds good, and thanks for the follow on issue for cache dropping.

> but the more I look at it, the more I think it should stay where it is [in arrow-benchmarks-ci], and we just cut out labs/benchmarks in the middle

I very much agree with this! If things like this are better off in arrow-benchmarks-ci there's nothing wrong with putting them there. But getting rid of the awkward dependency on labs/benchmarks in the middle is 💯

PR #123, opened 2022-12-21
Fix parentheses/quotation in documents

uribo:
I found this useful package and tried the examples in the README. The following error occurred when executing the code listed in the `A (contrived) example` section.

```
Error: unexpected ')' in:
"                     as_data_frame = c(TRUE, FALSE),
                     skip_empty_rows = TRUE))"
```

This is due to the extra parentheses, which are fixed in this PR.
In addition, I have also corrected a forgotten closing of parentheses in the Rd.

### DIFF
--- a/R/util.R
+++ b/R/util.R
@@ -72,7 +72,7 @@ bm_run_cache_key <- function(name, ...) {
 
 #' Confirm that the memory allocator enabled
 #'
-#' @param mem_alloc the memory allocator to be tested (one of: "jemalloc", "mimalloc", "system)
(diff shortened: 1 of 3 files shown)


jonkeane:
Thanks for the PR! Those are indeed typos. I'll merge this when the CI is green (not that I expect these changes would make it go red, but just in case there's something up with roxygen)

jonkeane:
Ha, turns out all that ❌  in CI is related to the new purr and not this PR: 124

PR #125, opened 2023-01-03
Deal with a purrr deprecation

jonkeane:
Adds `as.list()` (as suggested by the warning). The CI failures should all be resolved (and only are there because we depend on warnings being bubbled up)

### DIFF
--- a/R/util.R
+++ b/R/util.R
@@ -22,7 +22,7 @@ uuid <- function() {
 #' @importFrom stats setNames
 #' @importFrom rlang is_missing
 get_default_args <- function(FUN) {
-  forms <- formals(FUN)
+  forms <- as.list(formals(FUN))
   # Don't keep any of the formals that are length 0 (e.g. NULL)
   non_zero <- map_int(forms, length) > 0
   # Don't keep any of the formals that are missing (which happens if there is no default)


alistaire47 (approved):
If I understand correctly, this coerces the pairlist to a list so the `NULL`s become explicit before mapping, correct?

alistaire47:
ah wait, I can't comment on this line, but I think this makes the `missing` check line 29 unnecessary because missing become explicit `NULL`

jonkeane:
> ah wait, I can't comment on this line, but I think this makes the missing check line 29 unnecessary because missing become explicit NULL

Sneaky! Lemme try and see. When I added that missing check I definitely added some tests to confirm it works

alistaire47:
Wait, I'm wrong; it doesn't coerce to NULL, it keeps the empty symbol, so I think we do still need that. It literally just changes the class from pairlist to list.

alistaire47:
Odd that missing symbols are of length 1. I guess it's viewed more as an empty string (or NA? but not really, since it can't be held in an atomic vector) than NULL.

But sorry for the deviation; this should be good as-is.

jonkeane:
_nods_ before I saw those comments there, I added a breakpoint and confirmed exactly what you're saying too:

Where `forms` is `as.list(formals(FUN))`

```
Browse[1]> str(formals(FUN))
Dotted pair list of 5
 $ one     : num 1
 $ a_few   : language c(1, 2, 3)
 $ null    : NULL
 $ a_vector: symbol known_sources
 $ none    : symbol 
Browse[1]> str(forms)
List of 5
 $ one     : num 1
 $ a_few   : language c(1, 2, 3)
 $ null    : NULL
 $ a_vector: symbol known_sources
 $ none    : symbol 
```

jonkeane:
I'll merge when the CI is green

PR #131, opened 2023-02-02
Add benchmarks.json file for arrow-benchmarks-ci

alistaire47:
Adds benchmarks.json metadata file (and a little script to regenerate it if necessary) so arrow-benchmarks-ci can see what benchmarks are available in this repo.


jonkeane (approved):
Looks good.

PR #133, opened 2023-02-07
Let `run_id` be manually specified in runners

### LINKED ISSUE #132: Accept run_id in run.BenchmarkDataFrame
If we want to pass in a `run_id` from arrow-benchmarks-ci, `run.BenchmarkDataFrame` needs to accept a pre-specified one. It can default to `NULL`, which will result in one getting generated by benchconnect (the current behavior).

alistaire47:
Closes #132 by adding a `run_id` parameter to `run.BenchmarkDataFrame` and cascading it through the runner functions like `run_name` and `run_reason`. Will let arrow-benchmarks-ci specify `run_id`, but if a user calls `run(my_bm_df)` without specifying it, it will continue to generate one via benchconnect.


jonkeane (approved):
Looks good, thanks!

PR #135, opened 2023-06-08
dependencies: pin duckdb CRAN pkg to 0.7.1-1

jgehrcke:
From commit msg:
```
In an attempt to address
https://github.com/apache/arrow/issues/35917
https://github.com/voltrondata-labs/arrow-benchmarks-ci/issues/131

Assumption: the duckdb package move from
0.7.1-1 to 0.8.0 is responsible for result
differences. This patch will help to validate
or invalidate this assumption.

Let's see if this specific version pin breaks the
overall dependency resolution. If it works, we should stay here and have a benchmark owner/author do the investigation for what's required to perform a
duckdb bump.
```

https://github.com/apache/arrow/issues/35917
https://github.com/voltrondata-labs/arrow-benchmarks-ci/issues/131

I did read https://r-pkgs.org/description.html#sec-description-imports-suggests-minium-version to find the pinning syntax. Of course there's the warning

> Since R can’t have multiple versions of the same package loaded at the same time, specifying an exact dependency dramatically increases the chance of conflicting versions[2](https://r-pkgs.org/description.html#fn2).

but that always holds true. The result is what counts, and we will have to try it out to know the result :).


austin3dickey:
Does `duckdb (<= 0.7.1-1)` work?

I appreciate the quick temporary fix! If we can get it to work, I'd say we ship it and immediately start looking for a better solution. Version ceilings in R are even more fraught than in python.

jgehrcke:
> Does duckdb (<= 0.7.1-1) work?

pushed a commit to try that!

jonkeane:
_nods_ yeah this is what @alistaire47 was mentioning about pinning in the R ecosystem. 

I'll take a look at the underlying issue and see what I can do.

jgehrcke:
Closing, @jonkeane helped via https://github.com/voltrondata-labs/arrowbench/pull/136

PR #138, opened 2023-06-13
Return errored results

alistaire47:
Necessary for addressing https://github.com/voltrondata-labs/benchmarks/issues/134; a separate (coming) labs/benchmarks PR will be necessary to pick up all the data this is now sending properly.

The prime focus of this PR is to catch errors during iterations properly and write the results just like we do for non-errored benchmarks. Along the way this addresses some issues this uncovered, and makes a few judgment calls about what goes where, so comments are welcome!

### DIFF
--- a/R/measure.R
+++ b/R/measure.R
@@ -45,6 +45,7 @@ with_profiling <- function(profiling_on, expr) {
 }
 
 with_gc_info <- function(expr) {
+  force(expr)
   with_gcinfo <- "bench" %:::% "with_gcinfo"
   gc_output <- with_gcinfo(eval.parent(expr))
   # This will swallow errors, so check for error output and re-raise
--- a/R/run.R
+++ b/R/run.R
@@ -301,15 +301,19 @@ run_one <- function(bm,
       read_only = read_only,
       run_id = run_id,
       run_name = run_name,
+      batch_id = batch_id,
       run_reason = run_reason
     ),
     keep.null = TRUE
   )
   result <- do.call(run_script, run_script_args)
-  # If JSON reading fails because `run_bm()` errored, `batch_id` will not be populated
-  if (is.null(result$batch_id)) {
-    result$batch_id <- batch_id
+
+  if (!identical(Sys.getenv("TESTTHAT"), "true") && !is.null(result$error)) {
+    cat("Result errored; quitting with status = 1")
+    cat(result$error)
+    quit(status = 1L)
   }
+
   result
 }
 
@@ -339,12 +343,28 @@ run_bm <- function(bm, ..., n_iter = 1, batch_id = NULL, profiling = FALSE,
   })
 
   results <- list()
+  error <- NULL
   for (i in seq_len(n_iter)) {
-    results[[i]] <- run_iteration(
-      bm = bm,
-      ctx = ctx,
-      profiling = profiling,
-      drop_caches = global_params[["drop_caches"]]
+    tryCatch(
+      {
+        results[[i]] <- run_iteration(
+          bm = bm,
+          ctx = ctx,
+          profiling = profiling,
+          drop_caches = global_params[["drop_caches"]]
+        )
+      },
+      error = function(e){
+        # Note: this will only capture the last error
+        error <<- list(
+          error = as.character(e),
+          stack_trace = vapply(
+            traceback(3),
+            function(x) paste(x, collapse = "\n"),
+            character(1)
+          )
+        )
+      }
     )
   }
 
@@ -373,6 +393,18 @@ run_bm <- function(bm, ..., n_iter = 1, batch_id = NULL, profiling = FALSE,
     n_iter = n_iter
   )
 
+  if (length(result_df$real) == 0) {
+    stats <- NULL
+  } else {
+    stats <- list(
+      data = as.list(result_df$real),
+      unit = "s",
+      times = list(),
+      time_unit = "s",
+      iterations = n_iter
+    )
+  }
+
   out <- BenchmarkResult$new(
     run_name = run_name,
     run_id = run_id,
@@ -380,14 +412,8 @@ run_bm <- function(bm, ..., n_iter = 1, batch_id = NULL, profiling = FALSE,
     run_reason = run_reason,
     # let default populate
     # timestamp = utc_now_iso_format(),
-    stats = list(
-      data = as.list(result_df$real),
-      unit = "s",
-      times = list(),
-      time_unit = "s",
-      iterations = n_iter
-    ),
-    error = NULL,
+    stats = stats,
+    error = error,
     validation = NULL,
     tags = metadata$tags,
     info = metadata$info,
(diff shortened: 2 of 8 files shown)


R/measure.R
```diff
@@ -45,6 +45,7 @@ with_profiling <- function(profiling_on, expr) {
 }
 
 with_gc_info <- function(expr) {
+  force(expr)
```
alistaire47, line 48: This is necessary because `bench::with_gcinfo()` uses `sink()`, which does strange things with unevaluated promises—it swallows the error and keeps going, even though we have code to prevent exactly that.

R/run.R
```diff
@@ -301,15 +301,19 @@ run_one <- function(bm,
...
   result <- do.call(run_script, run_script_args)
-  # If JSON reading fails because `run_bm()` errored, `batch_id` will not be populated
-  if (is.null(result$batch_id)) {
-    result$batch_id <- batch_id
+
+  if (!identical(Sys.getenv("TESTTHAT"), "true") && !is.null(result$error)) {
+    cat("Result errored; quitting with status = 1")
+    cat(result$error)
+    quit(status = 1L)
   }
```
alistaire47, line 315: Previously `run_script()` would exit with status = 1, but `run_one()` (which is what labs/benchmarks calls) does not. This needs to get picked up by labs/benchmarks and escalated; hopefully that will make it through the conbench CLI.

R/run.R
```diff
@@ -339,12 +343,28 @@ run_bm <- function(bm, ..., n_iter = 1, batch_id = NULL, profiling = FALSE,
...
+        # Note: this will only capture the last error
+        error <<- list(
+          error = as.character(e),
+          stack_trace = vapply(
+            traceback(3),
+            function(x) paste(x, collapse = "\n"),
+            character(1)
+          )
+        )
+      }
```
alistaire47, line 367: Previously we were handling errors higher up, but here seems like the right place—if an iteration errors out, it should get caught, recorded, and escalated.

alistaire47: Also open to putting/naming what goes in `error` differently; `result.error.error` is literally the string of the error, which is hard to see from the traceback. Maybe we should switch to `rlang::trace_back()` or something?

R/run.R
```diff
@@ -617,11 +640,15 @@ run_script <- function(lines, cmd = find_r(), ..., metadata, progress_bar, read_
         output = NULL,
         rscript = lines
       ),
+      machine_info=NULL,
       context = metadata$context,
       github = metadata$github
     )
   }
 
+  result <- augment_result(result)
```
alistaire47, line 649: We could drop this. It make the results passed back to labs/benchmarks more fully-featured, but it repopulates most of that anyway; when arrowbench is called through abci directly, it will use the higher-level benchconnect run workflow.

R/util.R
```diff
@@ -64,6 +64,7 @@ file_with_ext <- function(file, new_ext) {
 
 bm_run_cache_key <- function(name, ...) {
   dots <- list(...)
+  dots <- lapply(dots, function(x) if (is.null(x)) "NULL" else x)
```
alistaire47, line 67: This was making dumb names like `results/placebo/character(0).json`

tests/testthat/test-util.R
```diff
@@ -41,10 +41,10 @@ test_that("sync_and_drop_caches() works", {
     }
   }
 
-  cases = purrr::cross(list(
+  cases = suppressWarnings(purrr::cross(list(
```
alistaire47, line 44: `purrr::cross()` is depressingly [getting deprecated](https://github.com/tidyverse/purrr/issues/768). I looked at alternatives, but they involve adding deps and going through dataframes for what should be simple, so I'll probably rewrite a version from scratch whenever this starts actually failing. Till then, this keeps the tests passing.

R/run.R
```diff
@@ -339,12 +343,28 @@ run_bm <- function(bm, ..., n_iter = 1, batch_id = NULL, profiling = FALSE,
...
+      {
+        results[[i]] <- run_iteration(
+          bm = bm,
+          ctx = ctx,
+          profiling = profiling,
+          drop_caches = global_params[["drop_caches"]]
+        )
+      },
+      error = function(e){
+        # Note: this will only capture the last error
```
austin3dickey, line 358: Should this be a list of all errors?

austin3dickey: oh right it should be something mappable to `{...}` json. Maybe `{"errors": [{...}]}`?

austin3dickey (approved):
thank you! Excited to see what happens next week

R/run.R
alistaire47: Reverting; this causes more problems than it fixes. Having nice exit statuses is important for CLIs, but not for libraries where we have more sophisticated error handling already.

R/run.R
alistaire47: > it should be something mappable to `{...}` json.

Right, it just has to be a dict, but it's schemaless. I took a look at other implementations, and a `stack_trace` key is pretty common, so I used that. But the R traceback doesn't always clearly identify what the error really is, annoyingly, so I put the stringified version of the error in as a separate `error` key, though this does mean you have `result.error.error`, which is a little unappetizing.

> Maybe `{"errors": [{...}]}`?

There's no need for a list here because when `tryCatch` hits an error in `expr`, it throws to the `error` handler (if specified), and it exits after it executes that—it doesn't restart where it left off. So 

``` r
> tryCatch({stop("a"); stop("b")}, error = identity)
<simpleError in doTryCatch(return(expr), name, parentenv, handler): a>
```
never hits `stop("b")`, so you'll always have one error. You can restart with `withCallingHandlers()`, but here that would make it more confusing where things actually went wrong.

This capturing is not very good at reproducing what you see when thing blow up in a complicated manner, though; it doesn't capture warnings (I think I might rearrange to add that, though they'll be under a separate `warnings` key instead of interleaved) or messages or other conditions (not sure that's worth adding), and if compiled code blows up, I'm not sure what will happen (though if it's a segfault the whole R session will exit).

tl;dr I'm aiming at

``` json
{
  ...
  "error": {
    "error": "stringified error",
    "stack_trace": ["traceback", "lines"],
    "warnings": ["optional", "array", "of", "warnings"]
  },
  ...
}
```
though I'm open to alternatives.

alistaire47: > Maybe `{"errors": [{...}]}`

Ohhhh, I think I get the intent of this now; do you mean it's a list with an element for each iteration? We could do that (instead of just capturing from the last iteration), but it's going to get very verbose and duplicative

R/run.R
alistaire47: Reverted; there are just not enough safeguards at this point in `run_one()` that `run_id` etc. are set correctly. This still works through the higher-level interface.

alistaire47:
Ok, I think this is polished and ready to go, but holding off on merging till [the accompanying changes to labs/benchmarks](https://github.com/voltrondata-labs/benchmarks/tree/edward/pass-through-r-errors) are finished in case I need any other tweaks here.

PR #139, opened 2025-06-25
Fix for failing R benchmarks - cast all decimal32/64 to decimal128

thisisnic:
Benchmarks are failing (https://github.com/apache/arrow/issues/46716) because decimal32 and decimal64 types have been implemented in Arrow but not Parquet, so when this type is being written to disk there is an error. This PR upcasts any decimal32/64 instances to decimal128 when writing to Parquet in R.

Haven't been able to test this out locally, so I've gone for a bit of an overkill approach and used this in calls to `write_parquet()`.

### DIFF
--- a/R/bm-write-file.R
+++ b/R/bm-write-file.R
@@ -25,7 +25,7 @@ write_file <- Benchmark("file-write",
       if (file_type == "feather") {
         return(function(...) arrow::write_feather(..., compression = compression))
       } else if (file_type == "parquet") {
-        return(function(...) arrow::write_parquet(make_table_parquet_compatible(...), compression = compression))
+        return(function(...) arrow::write_parquet(..., compression = compression))
       } else if (file_type == "fst") {
         # fst is always zstd, just a question of what level of compression
         level <- ifelse(compression == "uncompressed", 0, 50)
--- a/R/bm-write-file.R
+++ b/R/bm-write-file.R
@@ -25,7 +25,7 @@ write_file <- Benchmark("file-write",
       if (file_type == "feather") {
         return(function(...) arrow::write_feather(..., compression = compression))
       } else if (file_type == "parquet") {
-        return(function(...) arrow::write_parquet(..., compression = compression))
+        return(function(...) arrow::write_parquet(make_table_parquet_compatible(...), compression = compression))
       } else if (file_type == "fst") {
         # fst is always zstd, just a question of what level of compression
         level <- ifelse(compression == "uncompressed", 0, 50)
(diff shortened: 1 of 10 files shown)


R/bm-write-file.R
```diff
@@ -25,7 +25,7 @@ write_file <- Benchmark("file-write",
       if (file_type == "feather") {
         return(function(...) arrow::write_feather(..., compression = compression))
       } else if (file_type == "parquet") {
-        return(function(...) arrow::write_parquet(..., compression = compression))
+        return(function(...) arrow::write_parquet(make_table_parquet_compatible(...), compression = compression))
```
jonkeane: I don't think we should add `make_table_parquet_compatible` here. There shouldn't be a huge amount of overhead, but we don't want to add to our writing benchmark with additional type checking (and casting). IIUC, none of our write benchmarks are failing, but if they start to due to this, we should update the data we're using as an input rather than cast during the benchmark.

R/ensure-format.R
```diff
@@ -138,7 +138,7 @@ get_write_function <- function(format, compression, chunk_size = NULL) {
   if (format == "feather") {
     return(function(...) arrow::write_feather(..., chunk_size = chunk_size %||% 65536L, compression = compression))
   } else if (format == "parquet") {
-    return(function(...) arrow::write_parquet(..., chunk_size = chunk_size, compression = compression))
+    return(function(...) arrow::write_parquet(make_table_parquet_compatible(...), chunk_size = chunk_size, compression = compression))
```
jonkeane: I suspect that this isn't necessary (since it's just the TPC-H benchmarks that are failing and the changes in `custom-duckdb.R` should cover that, it's not awful to have an additional guard here when making these files.

jonkeane (approved):
Thanks for this. This looks good (though we should remove the addition in `bm-write-file.R`)

thisisnic:
Cheers @jonkeane, have now removed a couple of the additions there.

jonkeane:
😢 Apparently I can't approve the CI runs, nor merge to this repo. @assignuser could you assist?

assignUser:
I don't have write access either, I'll request that asap

assignUser:
Ugh, looks like there are multiple issues that need to be fixed unrelated to the PR... I assume the env in the actual benchmark runner is pinned better than here?

thisisnic:
> Ugh, looks like there are multiple issues that need to be fixed unrelated to the PR... I assume the env in the actual benchmark runner is pinned better than here?

Unsure sorry.  Mind approving the workflows again so I can check my changes worked for at least the tests bit for now?

